PR #154, opened 2026-04-20
fix: add WithDiscardUnknownFields client and call options for forward-compatible JSON unmarshaling

### LINKED ISSUE #153: protoc-gen-go-client: use DiscardUnknown for forward-compatible JSON unmarshaling
## Problem

The generated HTTP clients use bare `protojson.Unmarshal(body, msg)` which rejects unknown fields. When an API returns a field not defined in the proto schema, the entire request fails:

```
failed to unmarshal response: proto: (line 1:45): unknown field "swap_rate"
```

## Solution

Generated clients now include runtime options to opt into forward compatibility:

```go
// Service-level: all RPCs discard unknown fields
client := services.NewMarketDataServiceClient(baseURL,
    services.WithMarketDataServiceDiscardUnknownFields(true),
)

// Per-call: override for a single RPC
resp, err := client.GetQuote(ctx, req,
    services.WithMarketDataServiceCallDiscardUnknownFields(true),
)
```

**Default behavior is unchanged** (strict mode). Per-call options override client-level options. No breaking changes — the variadic `opts ...CallOption` parameter already exists on all generated RPC methods.

## Why it matters

APIs evolve — providers add new response fields regularly. Without a way to opt into forward compatibility, every new undocumented field breaks SDK users at runtime.

shavaizknz:
## Summary

- Generated clients now include `WithXxxDiscardUnknownFields(bool)` (service-level) and `WithXxxCallDiscardUnknownFields(bool)` (per-call) options
- When enabled, unknown fields in JSON responses are silently discarded instead of causing errors
- Default behavior is unchanged (strict mode) — **no breaking change**

## Usage

```go
// Service-level: all RPCs discard unknown fields
client := services.NewMarketDataServiceClient(baseURL,
    services.WithMarketDataServiceDiscardUnknownFields(true),
)

// Per-call: override for a single RPC
resp, err := client.GetQuote(ctx, req,
    services.WithMarketDataServiceCallDiscardUnknownFields(true),
)
```

### Precedence

| Client-level | Per-call | Result |
|---|---|---|
| not set | not set | strict (default) |
| `true` | not set | discard unknown |
| not set | `true` | discard unknown |
| `true` | `false` | strict (per-call wins) |

## What changed

- `internal/clientgen/generator.go` — added `discardUnknownFields bool` to client struct, `*bool` to call options struct, generated option functions, and runtime resolution in every RPC method
- Golden files regenerated (all existing RPCs pass `false` by default, option functions added)
- Tests and example updated

## Generated code

```go
// Client struct gets a new field
type marketDataServiceClient struct {
    ...
    discardUnknownFields bool
}

// Service-level option
func WithMarketDataServiceDiscardUnknownFields(discard bool) MarketDataServiceClientOption { ... }

// Per-call option (pointer to distinguish "not set" from "set to false")
type marketDataServiceCallOptions struct {
    ...
    discardUnknownFields *bool
}
func WithMarketDataServiceCallDiscardUnknownFields(discard bool) MarketDataServiceCallOption { ... }

// Runtime resolution in every RPC method
discardUnknown := c.discardUnknownFields
if callOpts.discardUnknownFields != nil {
    discardUnknown = *callOpts.discardUnknownFields
}
```

## Test plan

- [x] `TestForwardCompatibility` — 6 subtests proving DiscardUnknown handles all JSON types
- [x] `TestStrictModeRejectsUnknownFields` — confirms default strict behavior
- [x] `TestGeneratedCodeHasDiscardUnknownFieldsOption` — verifies golden file has client option, call option, and override logic
- [x] All 14 golden file tests pass
- [x] `go test ./...` — full suite passes (all 8 packages)
- [x] `go run ./examples/forward-compatibility/` — demonstrates the feature

Closes #153


AliMsayleb:
@shavaizknz to simplify it, can we do the client vs method support in two PRs ? Let's prioritize the client one for now, keep it small and simple, and then we'll do the method one as a followup later. What do you think ?

shavaizknz:
The per-call option is only ~10 lines on top of the client-level one (a `*bool` field, one option func, and a 3-line nil check). Both touch the same files and regenerate the same golden files, so splitting wouldn't meaningfully reduce the PR size. Happy to split if you feel strongly, but IMO it's simpler to land together since they're tightly coupled.

SebastienMelki (changes requested):
Thanks for the PR. The client-level/per-call API shape looks good, and the `*bool` call option is the right way to preserve "unset" vs "explicit false". I do not think this is mergeable yet though.

1. CI is red on lint. `Lint & Code Quality` currently fails on the new example plus one long generated-generator line. The new `examples/forward-compatibility` directory is the only Go example without its own `go.mod`, so it gets linted under the root module's stricter rules. Please either isolate it like the other examples or make it pass root lint, and wrap the long `generator.go` line.

2. `DiscardUnknown` is bypassed for generated custom JSON response types. `unmarshalResponse` returns through `json.Unmarshaler` before reaching the new `discardUnknown` branch, while generated helpers such as bytes/nullable/int64/flatten/oneof/timestamp handling still call bare `protojson.Unmarshal`. That means opt-in clients still reject unknown fields for many annotated response types. This needs a generated path that preserves the custom decoding behavior while using `protojson.UnmarshalOptions{DiscardUnknown: true}` when requested.

3. SSE streaming ignores the option entirely. `EventStream.Next` hard-codes `protojson.Unmarshal([]byte(data), event)`, and streaming RPCs never thread the resolved client/call `discardUnknown` value into the stream. Since SSE payloads have the same forward-compatibility problem, please either make `EventStream` honor the option or explicitly scope/document this feature as non-streaming only.

4. The tests do not currently exercise the generated client behavior. `forward_compat_test.go` mostly tests `protojson` directly and greps a golden file. Please add an end-to-end generated-client test with an `httptest.Server` returning an unknown JSON field, covering default strict mode, client-level discard, and per-call override back to strict. Ideally include one custom-JSON response path and SSE if SSE is supported.

The strict `false` in `handleErrorResponse` looks intentional to avoid loose JSON falsely matching `ValidationError`/`Error`; a short generator comment there would help preserve that intent.

shavaizknz:
Thanks for the detailed review. Here's where we are:

**1. CI lint — Fixed.** Added `go.mod` to `examples/forward-compatibility/` so it's isolated like the other examples. Wrapped the long `generator.go` line. `make lint` passes with 0 issues.

**3. SSE streaming — Fixed.** Added `discardUnknownFields bool` to the `EventStream` struct. `Next()` now conditionally uses `protojson.UnmarshalOptions{DiscardUnknown: true}` based on this field. The resolved flag (client default + per-call override) is threaded into the stream at creation time in `generateSSERPCMethod`.

**4. End-to-end tests — Fixed.** Added `TestUnmarshalResponseBehavior` with an `httptest.Server` returning unknown JSON fields, covering strict mode, client-level discard, and per-call override back to strict. Also added `TestSSEGoldenHasDiscardUnknownFields` to verify the SSE golden file has the correct wiring. Additionally, the `examples/forward-compatibility/` is now a full end-to-end example with a proto service, generated `QuoteServiceClient`, and a mock server — demonstrating all three modes using the actual generated client (no raw `protojson` usage).

**5. `handleErrorResponse` comment — Fixed.** Added a comment explaining why error parsing always uses strict `false` to avoid loose JSON falsely matching `ValidationError`/`Error`.

**2. Custom unmarshalers bypassing `DiscardUnknown` — Proposed solution:**

The challenge is that `json.Unmarshaler` is a standard library interface with the signature `UnmarshalJSON([]byte) error` — we don't own it and can't add parameters to it. The custom unmarshalers we generate (for int64, nullable, flatten, etc.) must implement this interface so they work with `encoding/json`. And the message structs themselves are generated by protoc-gen-go, not by us, so we can't add a `discardUnknown` field to carry the flag through.

Since we generate both sides (the caller in `unmarshalResponse` and the custom methods), we can define our own interface that carries the flag:

```go
// Generated once per client file
type discardAwareUnmarshaler interface {
    UnmarshalJSONWithDiscard([]byte, bool) error
}
```

In `unmarshalResponse`, check for this interface first:

```go
if unmarshaler, ok := msg.(discardAwareUnmarshaler); ok {
    return unmarshaler.UnmarshalJSONWithDiscard(body, discardUnknown)
}
if unmarshaler, ok := msg.(json.Unmarshaler); ok {
    return unmarshaler.UnmarshalJSON(body)
}
```

Each custom unmarshaler (int64, nullable, flatten, etc.) implements the new method and uses `protojson.UnmarshalOptions{DiscardUnknown: discard}` at the end instead of bare `protojson.Unmarshal`. The existing `UnmarshalJSON` stays for backward compat and delegates with `false`.

Would this approach work for you?

shavaizknz:
Review item 2 (custom unmarshalers bypassing DiscardUnknown) is addressed in a follow-up branch: shavaizknz/sebuf#1 (targeting `fix/discard-unknown-fields`). Once this PR merges, that branch can be PR'd to upstream.

Alternatively, if you could add me as a collaborator on the repo, I can create stacked PRs directly against upstream branches — makes the review flow faster since you can see clean diffs between each PR without cross-fork references.

AliMsayleb:
https://github.com/SebastienMelki/sebuf/pull/163

Solved in here

PR #159, opened 2026-04-23
fix: thread DiscardUnknown through custom JSON unmarshalers

shavaizknz:
## Summary

Addresses review item 2 from #154: custom `UnmarshalJSON` methods bypass the `DiscardUnknownFields` option.

**Depends on #154** — this branch builds on top of `fix/discard-unknown-fields`. Please review and merge #154 first, then this PR.

## Problem

Custom `UnmarshalJSON` methods (int64 encoding, nullable, flatten, oneof, timestamp, bytes, empty behavior) call bare `protojson.Unmarshal`, bypassing the `DiscardUnknownFields` option. When a response type uses any of these annotations, unknown fields still cause errors even with `WithXxxDiscardUnknownFields(true)`.

## Solution

Since `json.Unmarshaler` is a stdlib interface we can't modify, and message structs are owned by protoc-gen-go, we generate a new method on each custom type:

```go
// Thin wrapper for backward compat
func (x *Int64EncodingTest) UnmarshalJSON(data []byte) error {
    return x.UnmarshalJSONWithDiscard(data, false)
}

// Actual logic with discard support
func (x *Int64EncodingTest) UnmarshalJSONWithDiscard(data []byte, discardUnknown bool) error {
    // ... existing field transformations ...
    if discardUnknown {
        opts := protojson.UnmarshalOptions{DiscardUnknown: true}
        return opts.Unmarshal(modified, x)
    }
    return protojson.Unmarshal(modified, x)
}
```

In `unmarshalResponse`, an anonymous interface assertion checks for the new method first:

```go
if unmarshaler, ok := msg.(interface{ UnmarshalJSONWithDiscard([]byte, bool) error }); ok {
    return unmarshaler.UnmarshalJSONWithDiscard(body, discardUnknown)
}
// Fall back to json.Unmarshaler for types without custom encoding
if unmarshaler, ok := msg.(json.Unmarshaler); ok {
    return unmarshaler.UnmarshalJSON(body)
}
```

Applied to both `clientgen` and `httpgen` generators to maintain consistency.

## Test plan

- [x] All golden file tests pass (clientgen + httpgen)
- [x] Consistency tests pass (httpgen verifies clientgen and httpgen produce identical encoding)
- [x] `make lint` — 0 issues
- [x] `go test ./...` — all 8 packages pass


shavaizknz:
This PR depends on #154 (WithDiscardUnknownFields client/call options). Please review #154 first — this builds on top of it to fix custom unmarshalers that were bypassing the option.

Issue #171, opened 2026-05-19
protoc-gen-py-client: emit __init__.py re-exports


SebastienMelki:
Today consumers import directly from the generated file (`from note_service_client import NoteServiceClient`). Generating an `__init__.py` in each output package with `__all__` re-exports would let consumers write `from generated import NoteServiceClient` and would feel more like a real Python package.

Issue #170, opened 2026-05-19
protoc-gen-py-client: surface proto comments as Python docstrings


SebastienMelki:
Generated classes and methods currently carry a generic `"""Generated from proto message X."""` docstring. Real proto comments (leading + trailing) would render naturally as the dataclass / method docstring, improving IDE help text and Sphinx output.

Issue #169, opened 2026-05-19
protoc-gen-py-client: protobuf binary content-type support


SebastienMelki:
The Python client surfaces a `content_type` option on call options but raises `NotImplementedError` for anything other than `application/json`. Implementing `application/x-protobuf` would close feature parity with the Go and TS clients but currently requires a `google-protobuf` runtime dependency, which the generator explicitly avoids.

Decision pending:
- Ship as opt-in (plugin opt enables the binary path and emits the extra import), or
- Wait until a stdlib-only solution is feasible.

Issue #168, opened 2026-05-19
protoc-gen-py-client: async transport variant


SebastienMelki:
The generated client today only accepts a sync `HttpTransport` (returns `HttpResponse` directly). An async variant — returning `Awaitable[HttpResponse]` and generating `async def` RPC methods — would let consumers plug in `aiohttp` / `httpx.AsyncClient` natively.

Open questions:
- Emit both sync and async classes per service, or pick one per generator invocation via a plugin opt?
- Lean toward emitting both for ergonomic parity with Go's `Client` vs. TS's promise-returning surface.

PR #172, opened 2026-05-19
feat(py-client): add protoc-gen-py-client (Python HTTP client generator)

SebastienMelki:
Adds `protoc-gen-py-client`, the sixth generator in the sebuf toolkit. Generates type-safe Python HTTP clients that depend only on the Python standard library (Python 3.10+); users plug in `requests` / `httpx` / `aiohttp` via a duck-typed `HttpTransport` Protocol at construction time.

## What's in the box

- **`cmd/protoc-gen-py-client/`** + **`internal/pyclientgen/`** — full generator
- Per-proto-file output: message `@dataclass`es, `IntEnum`s, `HttpTransport` Protocol + `UrllibTransport` default, typed `*ClientOptions` / `*CallOptions`, one client class per service, and a per-`*Error`-message `ApiError` subclass hierarchy chosen at runtime from response shape via `_ERROR_CLASSES`
- All `(sebuf.http.*)` JSON-mapping annotations honored: `int64_encoding`, `enum_encoding` + `enum_value`, `nullable`, `empty_behavior`, `timestamp_format`, `bytes_encoding`, `oneof_config` + `oneof_value`, `flatten` + `flatten_prefix`, `unwrap` (map-value + root)
- Path params, query params (`urlencode(doseq=True)`), typed service + method header kwargs, content-type negotiation surface, Python 3.10 keyword escaping (hard + soft)
- 15 per-feature golden tests + helper unit tests; every generated file is fed to `python3 -c "import ast; ast.parse(...)"` so syntactic regressions are caught even when the golden string-compare wouldn't fire
- `examples/python-client-demo/` mirrors `examples/ts-client-demo/` section-by-section against the same Go HTTP server — `make demo` runs the full suite end-to-end

## What's deferred

Filed as follow-up issues:
- #167 SSE streaming (detection + `NotImplementedError` stub ships now; full iterator support is the follow-up)
- #168 Async transport variant
- #169 Protobuf binary content-type
- #170 Surface proto comments as Python docstrings
- #171 Emit `__init__.py` re-exports

## Generator bugs caught during Phase 2 testing

These were latent in the implementation commits and fixed before locking the goldens:
- Empty error-registry set emitted as `{}` (a dict) instead of `set()`
- Redundant `pass` followed by methods on field-less messages
- `google.protobuf.Timestamp` typed as `int`/`str` for unix/date formats while the encoder always called `datetime` methods — aligned on `datetime` everywhere
- Missing None-guard on WKT message-kind fields (Timestamp, Duration, Any, …) in `to_dict`
- Enum variant names were prefix-stripped and lowercased (`Priority.high`) but the encoder emitted `IntEnum.name`, producing `\"high\"` instead of Go's `\"PRIORITY_HIGH\"` — wire-format-breaking. Restored verbatim proto names

## Credit

This builds on test fixtures and serialization ideas from @elzalem's #132. The architecture intentionally follows the existing `tsclientgen` / `clientgen` pattern (calling `internal/annotations/` directly) rather than the `contractmodel` layer from that PR — so we could ship without blocking on the C# generator design discussion separately tracked on #131.

## Test plan

- [ ] `make build` — produces `bin/protoc-gen-py-client`
- [ ] `go test ./internal/pyclientgen/...` — golden tests + helper unit tests all pass
- [ ] `golangci-lint run ./internal/pyclientgen/... ./cmd/protoc-gen-py-client/...` — 0 issues
- [ ] `cd examples/python-client-demo && make demo` — all 7 demo sections green end-to-end against the real Go server
- [ ] Eyeball `docs/python-generation.md` and the new entries in `README.md` + `CLAUDE.md`


yashagarwal-sarwa (approved):

yashagarwal-sarwa:
Bug: error classes emitted before the enums they reference

When a proto message with an Error suffix references an enum field with a default value, the generated error class is declared before the enum — causing a NameError at import time.

Reproduction: Our EventError message has code: RejectionReason as a field. The generated output emits class EventError(ApiError) at line 99, but class RejectionReason(IntEnum) isn't declared until line 215.


class EventError(ApiError):
    code: RejectionReason = RejectionReason.REJECTION_REASON_UNSPECIFIED  # NameError here
Fix: Enums should be emitted before error classes in the output ordering. Looks like it's a topological sort issue in the emission order — error classes need to come after all enums they depend on.

SebastienMelki:
@yashagarwal-sarwa great catch — confirmed and fixed in 0166439.

Root cause was exactly what you diagnosed: `writeErrors` ran before the enum loop in `generateClientFile`, so any `*Error` message with an enum-typed field referenced a class that wasn't defined yet. `from __future__ import annotations` only stringifies *annotations*, not default-value expressions, which are evaluated at `def`-time.

The fix is a one-line reorder (enums before errors) — messages already trailed both blocks, and message-typed defaults are always `None` so they're safe. Added an `EventError` + `RejectionReason` regression case to `testdata/proto/errors.proto` matching the exact shape you reported.

Also upgraded the golden test: every generated file now goes through `importlib.util.spec_from_file_location` + `exec_module` (with the module registered in `sys.modules` so `@dataclass` can resolve forward annotations). `ast.parse` alone wouldn't have caught this — `NameError` is a runtime error, not syntactic — so the import check is the real backstop going forward.

Verified end-to-end against the python-client-demo Go server: clean run.

SebastienMelki:
@yashagarwal-sarwa fix is up in 0166439 (now on the branch). Mind taking another look when you have a minute?

yashagarwal-sarwa:
EventError.from_dict() missing — EventResult.from_dict() fails at runtime

Error classes generated from proto *Error messages get to_dict() and populate(), but NOT from_dict(). However, when EventError appears as a field on a regular message (EventResult.error), the generated EventResult.from_dict() calls EventError.from_dict() which doesn't exist:


# In EventResult.from_dict() — generated code:
kwargs["error"] = EventError.from_dict(data["error"])  # AttributeError: no from_dict
Fix: Either add from_dict() as an alias for populate() on error classes, or have the message deserializer use populate() when the target type is an error class.

SebastienMelki:
@yashagarwal-sarwa update — pushed the from_dict fix you flagged, plus broader coverage. The new commits are:

- **3521398** — your fix: from_dict alias for *Error classes
- **608ff0a** — three more generator bugs the comprehensive demo caught
- **53a7826** — two new end-to-end examples: `python-encoding-demo` (51 assertions, every JSON-mapping annotation) and `python-errors-demo` (41 assertions, every error surface). Both `make demo` from their dirs and round-trip against a real Go server.

The encoding demo specifically surfaced three issues we didn't have coverage for before:
1. `TIMESTAMP_FORMAT_DATE` decoded the wire string as a `str` even though the field type is `datetime` (annotation/code mismatch on the decode side)
2. flatten emitted wire keys with JSON-camelCase names while the Go HTTP plugin uses proto snake_case — so `Address.zip_code` shipped as `author_zipCode` from Python but the server expected `author_zip_code`
3. root-map unwrap silently produced empty `to_dict()` / `from_dict()` because `annotations.FindUnwrapField` is list-only by design (used by tsclientgen/openapiv3) and py-client called it on root unwrap paths

All fixed and round-tripping clean. Mind taking another look?

PR #179, opened 2026-06-01
feat(gen/go-http): expose protojson.MarshalOptions on server registration

### LINKED ISSUE #178: protoc-gen-go-http: expose protojson.MarshalOptions on server registration
## Problem

The generated HTTP servers hardcode `protojson.Marshal(msg)` with default options. There is no way for an application registering a server to configure response JSON behavior, which forces lossy wire output: proto3 scalar zero values are silently dropped.

A field like `bool no_new_orders = 14;` set to `false` is indistinguishable from "unset" in proto3 wire semantics and disappears from the response, breaking clients that need to observe the explicit `false`. The same applies to empty strings, zero numbers, and unset enums whenever the application wants them emitted.

Hardcoded call sites in `internal/httpgen/generator.go` (origin/main):

- Lines `740`, `748` (`writeJSONResponse`)
- Lines `997`, `1004` (`writeProtoMessageResponse`)
- Lines `1185`, `1191` (secondary response writer)
- Lines `1660`, `1676` (SSE event)
- Line `1843` (error marshaling)

There is also a `json.Marshaler` fast path at lines `408` / `416` of the generated `*_http_binding.pb.go` that bypasses `protojson` entirely — same shape as the `json.Unmarshaler` problem #163 had to work around for custom client unmarshalers.

## Solution

Mirror the client-side pattern from #163 on the server: surface `protojson.MarshalOptions` as a single `ServerOption` and route it through every generated marshal site (including the custom marshaler path).

```go
err := ipov1.RegisterIpoServiceServer(handler,
    ipov1.WithMux(mux),
    ipov1.WithMarshalOptions(protojson.MarshalOptions{EmitUnpopulated: true}),
)
```

Accepting the full `protojson.MarshalOptions` struct (rather than granular `WithEmitUnpopulated(true)` style setters) keeps sebuf forward-compatible with upstream `google.golang.org/protobuf` without having to enumerate every knob.

Required changes in `internal/httpgen/generator.go`:

1. **Extend `serverConfiguration`** (line `820`) with a `marshalOpts protojson.MarshalOptions` field.
2. **Emit a `WithMarshalOptions` `ServerOption` constructor** alongside `WithMux` / `WithErrorHandler` (line `848`).
3. **Replace hardcoded calls** at the 9 marshal sites with the configured options, e.g. `config.marshalOpts.Marshal(msg)`. This requires `config` to be in scope at each site — `genericHandler`, `writeProtoMessageResponse`, `writeJSONResponse`, the SSE handler, and `handleErrorResponse` all need plumbing changes similar to how `errorHandler` is already threaded.
4. **Custom marshaler interface (mirror of `sebufUnmarshaler` from #163)** — introduce `sebufMarshaler`:

    ```go
    type sebufMarshaler interface {
        MarshalJSONSebuf(opts protojson.MarshalOptions) ([]byte, error)
    }
    ```

    All 8 custom marshaler generators (`bytes_encoding`, `empty_behavior`, `encoding`, `flatten`, `nullable`, `oneof_discriminator`, `timestamp_format`, `unwrap`) emit `MarshalJSONSebuf` plus a thin `MarshalJSON` wrapper that calls it with default options for backward compatibility with `encoding/json` and the existing `json.Marshaler` fast path. Wrapper/flatten/oneof generators forward opts to nested children. The dispatcher in the HTTP binding checks `sebufMarshaler` first, then `json.Marshaler`, then `config.marshalOpts.Marshal`.

5. **Defaults unchanged** — zero-value `MarshalOptions{}` preserves current behavior. No breaking change for existing servers.
6. **Tests** — golden assertions that the options thread through every generated handler (standard, SSE, error response, custom marshaler path), plus httptest integration tests covering `EmitUnpopulated: true` emitting `false`.

## Why it matters

Server JSON encoding is currently rigid: applications can't surface `false`, empty strings, or zero numbers when those carry semantic meaning, and they can't tune any other `protojson` knob (`UseProtoNames`, `UseEnumNumbers`, `Multiline`, etc.) without forking sebuf. Exposing `protojson.MarshalOptions` at the registration boundary — the same boundary where `WithMux` and `WithErrorHandler` already live — is the natural mirror of #163 and keeps proto JSON semantics under the application's control instead of pinned to upstream defaults.

Related: #163 (client-side `UnmarshalOptions`)

AliMsayleb:
## Description
Generated HTTP servers hardcode `protojson.Marshal` with default options, so proto3 zero values silently drop from JSON responses — clients can't observe an explicit `false` on `bool no_new_orders` (or empty strings, zero numbers, unset enums).

This adds `WithMarshalOptions(protojson.MarshalOptions)` as a `ServerOption` and threads it through every generated marshal site, including the 8 custom marshaler generators that previously bypassed `protojson` via the `json.Marshaler` fast-path. Each annotated message now emits `MarshalJSONSebuf(opts) ([]byte, error)` plus a thin `MarshalJSON()` wrapper for back-compat. Dispatch order at the JSON boundary is `sebufMarshaler` → `json.Marshaler` → `protojson`.

Mirrors the client-side pattern from #163 on the server. Server-side `UnmarshalOptions` is explicitly out of scope.

## Related Issue
Fixes #178

## Type of Change
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
- [x] Unit tests pass (`go test ./...`)
- [x] Integration tests pass — new `TestMarshalOptionsIntegration` spins up an `httptest` server and asserts `EmitUnpopulated:true` emits `"success":false` while default opts preserve byte-identical output
- [x] Golden file tests updated — every golden in `internal/httpgen/testdata/golden/` regenerated; only scaffolding changes on the default path
- [x] Manual testing performed — verified same-package cross-file scenarios (`same_pkg_*`, `cross_int64_*`) compile cleanly in a real Go module
- [x] `make lint` → 0 issues

### Test Configuration
- Go version: 1.26.0
- OS: macOS 26.3.1 (arm64)
- Protoc version: libprotoc 34.1

## Checklist
- [x] `go fmt ./...` clean
- [x] Self-reviewed
- [x] Comments added where the why is non-obvious (sebufMarshaler interface, marshalJSONWithOpts dispatch order, inline child dispatch rationale)
- [x] No new warnings (`go vet`, `golangci-lint` clean)
- [x] Tests added that prove the feature works (`marshal_options_integration_test.go`)
- [x] Existing tests pass with my changes (7 pre-existing `TestGoGeneratorsProduceIdentical*` consistency failures remain — UnmarshalJSON divergence from #163, out of scope)

## Breaking Changes
None. Zero-value `MarshalOptions{}` preserves current wire output byte-for-byte. `MarshalJSON` wrappers still satisfy `json.Marshaler` for third-party consumers.

## Screenshots/Examples
```go
err := ipov1.RegisterIpoServiceServer(handler,
    ipov1.WithMux(mux),
    ipov1.WithMarshalOptions(protojson.MarshalOptions{EmitUnpopulated: true}),
)
// Now responses surface `"no_new_orders": false` instead of dropping the field.
```

## Additional Notes
- The `clientgen` mirror generators also get the `MarshalJSONSebuf` rename — purely to keep the `TestGoGeneratorsProduceIdentical*` consistency tests honest on the marshal side. No client-facing API change.
- Related: #163 (client-side `UnmarshalOptions`)


SebastienMelki:
## Review: solid change, no breakage — a few polish items before shipping

Reviewed the full diff, ran the suite, and verified correctness end-to-end. The design is sound: the `sebufMarshaler` indirection (`MarshalJSONSebuf(opts)` → `json.Marshaler` → `protojson`) is the right way to thread options through the custom marshalers, back-compat is byte-identical with `MarshalOptions{}`, and there's no recursion risk (`opts.Marshal` uses proto reflection, never re-invoking the custom methods).

**Verification done:**
- `TestMarshalOptionsIntegration` passes — proves `EmitUnpopulated:true` surfaces `"success":false` and default omits it.
- Confirmed the 7 `TestGoGeneratorsProduceIdentical*` failures are **pre-existing on `main`** (from #163's unmarshal-side divergence) — reproduced them identically in a clean `main` worktree. This PR introduces **zero** new failures and re-aligns the marshal side.
- `golangci-lint` on changed packages: 0 issues.
- Grepped every remaining `protojson.Marshal` in generated output — all 3 survivors are correct to leave (2 in the unmarshal round-trip path which must stay canonical, 1 in the `Error() string` method which isn't the HTTP wire path).

### Polish items before shipping

**1. Document `WithMarshalOptions` (the main gap).**
`docs/http-generation.md` (~L764) only lists `WithMux` under the `ServerOption` section. Add `WithMarshalOptions` there — and while we're in, `WithErrorHandler` is also missing. Suggested addition:

```go
// WithMarshalOptions configures protojson.MarshalOptions used for JSON responses
// (including SSE events and error bodies). Zero value preserves default behavior.
// Use EmitUnpopulated to surface zero-value fields, UseProtoNames for snake_case, etc.
func WithMarshalOptions(opts protojson.MarshalOptions) ServerOption

// WithErrorHandler installs a custom handler for error responses.
func WithErrorHandler(h ErrorHandler) ServerOption
```

A short prose example mirroring the PR's `EmitUnpopulated` snippet would also help (`README.md` server section + `docs/http-generation.md`).

**2. Add a worked example (optional but nice).**
There's an `examples/error-handler/` for `WithErrorHandler`. A small `examples/marshal-options/` (or a note in an existing example) showing `EmitUnpopulated` flipping a zero-value field on would make the feature discoverable. #163 didn't add a client-side example for `UnmarshalOptions`, so skipping this stays consistent — but it's the highest-value doc artifact for this feature.

**3. (Awareness, not a fix) SSE post-commit error event.**
`generator.go:~1837` emits `err.Error()` for the after-commit SSE error case and doesn't route through `marshalOpts`. This is pre-existing and an edge case (the normal `sender.Send` path *does* now honor opts), so fine to leave — just flagging so it's a conscious decision.

Nothing here blocks merge; items 1–2 are the "fully polished" delta.

AliMsayleb:
Thanks for the review. Addressed the two polish items in bb49565:

**1. Docs** — `docs/http-generation.md` `ServerOption` section now lists `WithMux`, `WithErrorHandler`, and `WithMarshalOptions` with signatures, and includes a worked example showing `EmitUnpopulated: true` surfacing zero-value fields.

**2. Worked example** — added `examples/marshal-options/`. Mirrors the structure of `examples/error-handler/`. `make demo` boots two servers from the **same handler** on `:8080` (defaults) and `:8081` (`EmitUnpopulated: true`); `make test` curls both and shows the wire diff side-by-side. Verified locally:

```
=== Default opts (port 8080) — zero values omitted ===
{}

=== EmitUnpopulated: true (port 8081) — zero values surfaced ===
{"acceptingOrders":false,"subscriberCount":0,"note":""}
```

**3. SSE post-commit error event** — left as-is per your note. It's the edge case where headers have already flushed and we can no longer route through `marshalJSONWithOpts` (the response is committed to plain text by then); the pre-commit path and every other SSE write does honor opts.

SebastienMelki (approved):
Approved — thanks @AliMsayleb for the quick turnaround on the polish items. 🙏

Re-verified the latest commit (`bb49565`):
- **Docs** — `ServerOption` section now covers `WithMux`, `WithErrorHandler`, and `WithMarshalOptions` with a worked `EmitUnpopulated` example. Accurate.
- **Example** — generated, built, and ran `examples/marshal-options/` end-to-end. Reproduced the wire diff exactly: `:8080` → `{}`, `:8081` (`EmitUnpopulated: true`) → `{"acceptingOrders":false, "subscriberCount":0, "note":""}`. Exercising the feature through a fully generated server is the strongest validation we could ask for.
- **SSE post-commit error event** — correctly left as-is; the reasoning is sound.
- CI green (Lint/Test/Coverage/Build/Integration), no new test failures (the 7 `ProduceIdentical*` are pre-existing from #163).

Merging now and will push a new release tag (`v0.14.0`, minor bump for the non-breaking `WithMarshalOptions` addition) once the merge lands.

PR #198, opened 2026-06-30
feat(ts): opt-in import_style=modules and oneof_style=discriminated for TS generators

hishamank:
## Summary
Adds two **orthogonal, opt-in** plugin options to both TypeScript generators (`protoc-gen-ts-client`, `protoc-gen-ts-server`), parsed from the protoc parameter string:

- **`import_style=modules`** — emit one canonical type module per proto file (`<proto>.ts`) with cross-file imports, plus a single shared `errors.ts`, instead of inlining the full type closure into every service file. Service files become slimmed clients/servers that import their request/response types. Fixes the type-duplication problem (e.g. `SongID`/`ValidationError` re-declared in every service file).
- **`oneof_style=discriminated`** — render un-annotated (non-synthetic) oneofs as `$case` discriminated unions instead of flattened optional fields. An explicit `sebuf.http` `oneof_config` annotation still wins.

Defaults are `import_style=inline` / `oneof_style=flatten`.

## Backward compatibility
With no options, output is **byte-for-byte identical** to today. This is enforced by the design: the existing emitter functions become thin wrappers that delegate to new `*Ctx` variants with a nil context; a nil/inline context resolves every type reference to its bare name and records no import. All existing goldens are unchanged (the new behavior lives in new golden subdirectories).

## ⚠️ `import_style=modules` requires `strategy: all`
The modules layout emits shared files (`errors.ts`, cross-package type modules). buf's default `strategy: directory` invokes the plugin once per directory, which would emit those shared files repeatedly. Set `strategy: all` on the plugin in `buf.gen.yaml` when using `import_style=modules`:
```yaml
plugins:
  - local: protoc-gen-ts-client
    out: ./web
    strategy: all
    opt:
      - import_style=modules
      - oneof_style=discriminated
```

## Implementation
- `internal/tscommon/config.go` — `Options` + `ParseOptions` (ignores unknown keys like `paths=source_relative`; errors on bad values).
- `internal/tscommon/imports.go` — `EmitContext` + `ImportTracker` (relative specifier computation, deterministic collision aliasing, type-only vs value imports).
- `internal/tscommon/modules.go` — `EmitSharedModules`: per-proto type modules grouped by source file + shared `errors.ts`; reserved-name guard (`ValidationError`/`ApiError`/`FieldViolation`).
- `internal/tscommon/types.go` — `*Ctx` variants of the emitters (nil-wrapper preserves defaults); `SynthesizeOneofInfo` (`$case`); `MessagesBySourceFile`/`EnumsBySourceFile`.
- Generators branch to modules mode and thread an `EmitContext` through type/enum reference sites; the shared type modules emitted by client and server are byte-identical.

## Tests / verification
- `go vet ./...` + full suite green (9 packages). Existing inline goldens unchanged.
- New golden cases (separate subdirs) for both generators: `import_style=modules` (per-proto module + slimmed service file + shared `errors.ts`) and `oneof_style=discriminated` (`$case` unions). Client/server shared type modules verified byte-identical.
- Unit tests: `ParseOptions`, `RelativeImportSpecifier`, `ImportTracker` (collision aliasing, render ordering).
- **End-to-end** against a real multi-package schema (anghamna, with `strategy: all`): 98 files, **0** duplicate-file warnings, `SongID` defined once and imported, oneofs discriminated, and the full tree type-checks under `tsc --strict --moduleResolution bundler`.

## Follow-ups (out of scope)
- A dedicated multi-package golden fixture (cross-package imports are covered by the e2e run for now).
- Docs page for the TS generator options.


hishamank:
Split into two independently-reviewable PRs so each feature can be accepted on its own:
- #199 — `oneof_style=discriminated` (smaller, self-contained)
- #200 — `import_style=modules` (stacked on #199)

Closing this combined PR in favor of those.

PR #211, opened 2026-07-20
feat(ts): emit protobuf-es client as standalone tree-shakeable functions

hishamank:
Stacks on **#210** (`feat/ts-es-result-errors`). Merge after #207 → #210.

_(Branch name `feat/ts-es-client-factory` is legacy — this PR now emits standalone functions, not a factory.)_

## What & why

The protobuf-es client was a `class` holding only immutable config (`baseURL`, `fetchFn`, `defaultHeaders`) set once. Rather than a client object at all, each RPC is now a **top-level `export async function`** that takes its config **per call**:

```ts
import { getUser } from "./gen/user_service_client.js";

const user = await getUser({ id: "123" }, { baseURL: "https://api...", headers: { apiKey } });
```

Config is a required `{Service}RequestOptions` argument (`baseURL` required; `fetch`, `headers`, `signal`, and typed header props optional).

- **Tree-shakeable** — every RPC is an independent top-level function, so bundlers drop the ones you don't import. Using one method no longer pulls the whole service into the bundle.
- **Stateless** — no client object to construct or thread around; hold the plain `RequestOptions` if you want to reuse config.
- **Composes with `ts_error_handling=result`** (#210) — standalone unary functions still return `Promise<Result<T, ClientError>>` via `decodeError`. `handleError` becomes a single top-level (unexported) helper per file, emitted only when referenced (SSE).
- **Hand-rolled mode unchanged** — still a configure-once `class` you `new`. Shared body emitters branch on a `recv()` prefix + a per-call config prologue; hand-rolled goldens are byte-identical.

## Verification

- `go build`, `go vet`, `staticcheck` clean; full `go test ./internal/...` green.
- Regenerated es + es-result goldens; **both** tsc `--noEmit` typecheck suites pass against `@bufbuild/protobuf@2.12.1`.


SebastienMelki (approved):
Thanks Hicham — clean transformation from class to standalone functions. The `recv()` prefix abstraction is elegant, tree-shaking is genuinely achievable with this output shape, and the Result + typed-headers composition both work correctly. Good to merge once #210 lands.

---

## Observations (no blocking issues)

### Design

- **Function signatures are correct**: `export async function getItem(req: MessageInitShape<...>, options: RequestOptions): Promise<Item>` — req first, required config second.
- **RequestOptions is well-designed**: `baseURL` required, `fetch`/`headers`/`signal` optional. Per-call config is the right trade-off for tree-shaking — no shared state means bundlers can eliminate unused functions.
- **Typed-header extends pattern works**: `TokenServiceRequestOptions extends RequestOptions` adds only the header properties from the proto annotation. Clean.
- **`recv()` abstraction** keeps the hand-rolled class path and es standalone path sharing the same body emitters with minimal branching.

### Correctness

- **Tree-shaking is real**: each RPC is an independent top-level function. Importing one does not transitively pull others (only shared `handleError` + protobuf-es schema imports come along).
- **handleError scoping is correct**: top-level unexported helper, emitted only when SSE methods reference it.
- **SSE as standalone `async function*`** is correct — `export async function* streamEvents(...)` with proper `AsyncGenerator<T>` return type.
- **Hand-rolled goldens are byte-identical** — no regressions.
- **Result mode composes cleanly**: standalone unary functions return `Promise<Result<T, ClientError>>` via `decodeError()`.

### Minor notes (non-blocking)

- **`baseURL` required per call** is the unavoidable cost of statelessness — consumers who want a configured client can wrap it themselves. Worth a one-liner in the docs suggesting a `const opts = { baseURL: "..." }` pattern for reuse.
- **Config prologue duplication** (2 lines per function: `baseURL.replace`, `options.fetch ?? globalThis.fetch`) is trivially small overhead and not worth abstracting.

This is the right architecture for es-mode — tree-shakeable, stateless, composable with Result and typed headers. Well done.

SebastienMelki:
Superseded by #242 (rebased onto main after squash-merge of the stacked PRs).

PR #212, opened 2026-07-20
fix(httpgen): emit UnmarshalJSONSebuf for unwrap and bytes encoding types

### LINKED ISSUE #204: [BUG] httpgen: unwrap types ignore DiscardUnknown
## Describe the bug

For a response message with `[(sebuf.http.unwrap) = true]`, `httpgen` generates `UnmarshalJSON` but not the options-aware `UnmarshalJSONSebuf` method.

The generated client creates `protojson.UnmarshalOptions{DiscardUnknown: true}` when configured with `With...DiscardUnknownFields(true)`. An unwrap response cannot receive those options because it only implements `UnmarshalJSON([]byte) error`. Its generated decoder then calls strict `protojson.Unmarshal`, so an unknown response field causes decoding to fail even when the HTTP status is 200.

Expected: unknown fields are ignored when `DiscardUnknownFields` is enabled.

Actual: the client returns an error such as `proto: (line 1:11): unknown field "unexpected_field"`.

Confirmed on v0.18.0, v0.19.0, v0.20.0, and main.

### 1. Source check on Windows

Run this from the sebuf repository root in PowerShell:

```powershell
$path = "internal/httpgen/unwrap.go"
$marshalCount = (Select-String -Path $path -Pattern "MarshalJSONSebuf").Count
$unmarshalCount = (Select-String -Path $path -Pattern "UnmarshalJSONSebuf").Count
"MarshalJSONSebuf=$marshalCount"
"UnmarshalJSONSebuf=$unmarshalCount"
```

Current output:

```text
MarshalJSONSebuf=14
UnmarshalJSONSebuf=0
```

`14` is the number of matching source lines, not the number of unwrap types.

### 2. Standalone HTTP response test

Create a temporary `_unwrap-repro` folder directly under a sebuf checkout. This module uses only the sebuf checkout, protobuf, and the Go standard library. It does not use Alpaca, alpaca-go, Sarwa code, credentials, a real server, or an existing reproduction package.

`_unwrap-repro/go.mod`:

```go.mod
module example.com/unwrap-repro

go 1.26.0

require (
	github.com/SebastienMelki/sebuf v0.0.0
	google.golang.org/protobuf v1.36.11
)

replace github.com/SebastienMelki/sebuf => ..
```

`_unwrap-repro/proto/example/v1/items.proto`:

```proto
syntax = "proto3";
package example.v1;

import "sebuf/http/annotations.proto";
option go_package = "example.com/unwrap-repro/gen/example/v1;examplev1";

message Item {
  string id = 1;
}

message ListItemsRequest {}

message ListItemsResponse {
  repeated Item items = 1 [(sebuf.http.unwrap) = true];
}

service ItemsService {
  option (sebuf.http.service_config) = {base_path: "/v1"};

  rpc ListItems(ListItemsRequest) returns (ListItemsResponse) {
    option (sebuf.http.config) = {
      path: "/items"
      method: HTTP_METHOD_GET
    };
  }
}
```

`_unwrap-repro/unwrap_discard_unknown_test.go`:

```go
package repro_test

import (
	"context"
	"io"
	"net/http"
	"strings"
	"testing"

	examplev1 "example.com/unwrap-repro/gen/example/v1"
)

type roundTripFunc func(*http.Request) (*http.Response, error)

func (f roundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
	return f(req)
}

func TestDiscardUnknownOnUnwrapResponse(t *testing.T) {
	transport := roundTripFunc(func(*http.Request) (*http.Response, error) {
		return &http.Response{
			StatusCode: http.StatusOK,
			Header: http.Header{
				"Content-Type": []string{"application/json"},
			},
			Body: io.NopCloser(strings.NewReader(
				`[{"id":"a","unexpected_field":"x"}]`,
			)),
		}, nil
	})

	client := examplev1.NewItemsServiceClient(
		"http://example.test",
		examplev1.WithItemsServiceHTTPClient(&http.Client{Transport: transport}),
		examplev1.WithItemsServiceDiscardUnknownFields(true),
	)

	resp, err := client.ListItems(context.Background(), &examplev1.ListItemsRequest{})
	if err != nil {
		t.Fatalf("DiscardUnknownFields should accept the response: %v", err)
	}
	if len(resp.GetItems()) != 1 || resp.GetItems()[0].GetId() != "a" {
		t.Fatalf("unexpected response: %+v", resp)
	}
}
```

From the sebuf repository root, build the generators and generate the standalone package:

```powershell
New-Item -ItemType Directory -Force bin | Out-Null
New-Item -ItemType Directory -Force _unwrap-repro/gen | Out-Null

go build -o bin/protoc-gen-go-http.exe ./cmd/protoc-gen-go-http
go build -o bin/protoc-gen-go-client.exe ./cmd/protoc-gen-go-client
go build -o bin/protoc-gen-go.exe google.golang.org/protobuf/cmd/protoc-gen-go
$env:PATH = "$(Resolve-Path ./bin);$env:PATH"

$protoc = "C:/path/to/protoc.exe"
& $protoc -I_unwrap-repro/proto -Iproto `
  --go_out=_unwrap-repro/gen --go_opt=paths=source_relative `
  --go-http_out=_unwrap-repro/gen --go-http_opt=paths=source_relative `
  --go-client_out=_unwrap-repro/gen --go-client_opt=paths=source_relative `
  _unwrap-repro/proto/example/v1/items.proto

Push-Location _unwrap-repro
go mod tidy
go test -run TestDiscardUnknownOnUnwrapResponse -count=1 -v .
Pop-Location
```

The test currently receives the in-memory HTTP 200 response and fails with `unknown field "unexpected_field"`. No network request is made because the custom `http.RoundTripper` returns the response directly.

## Root cause

The generated client checks the options-aware method first, then falls back to the standard JSON method:

```go
opts := protojson.UnmarshalOptions{DiscardUnknown: discardUnknown}

if u, ok := msg.(sebufUnmarshaler); ok {
	return u.UnmarshalJSONSebuf(body, opts)
}
if u, ok := msg.(json.Unmarshaler); ok {
	return u.UnmarshalJSON(body)
}
return opts.Unmarshal(body, msg)
```

The generated unwrap type only has this method:

```go
func (x *ListItemsResponse) UnmarshalJSON(data []byte) error {
```

Its message-element loop uses:

```go
if err := protojson.Unmarshal(itemRaw, item); err != nil {
```

`protojson.Unmarshal` is equivalent to `protojson.UnmarshalOptions{}.Unmarshal`, so `DiscardUnknown` is false there.

The source of both generated lines is `internal/httpgen/unwrap.go`, in `generateRootRepeatedUnwrapUnmarshalJSON`:

```go
gf.P("func (x *", msgName, ") UnmarshalJSON(data []byte) error {")
gf.P("if err := protojson.Unmarshal(itemRaw, item); err != nil {")
```

This is the generator bug. It emits only the fallback method and hardcodes a decoder that cannot receive the client options.

## Suggested fix

Generate `UnmarshalJSONSebuf` as the main implementation, use the supplied `opts` for protobuf children, and keep `UnmarshalJSON` as a compatibility wrapper with default options.

Representative source diff:

```diff
@@
-gf.P("func (x *", msgName, ") UnmarshalJSON(data []byte) error {")
+gf.P("func (x *", msgName, ") UnmarshalJSONSebuf(data []byte, opts protojson.UnmarshalOptions) error {")
@@
-if g.hasEncodingMarshalJSON(rootUnwrap.UnwrapField.Message) {
-	gf.P("if err := json.Unmarshal(itemRaw, item); err != nil {")
-} else {
-	gf.P("if err := protojson.Unmarshal(itemRaw, item); err != nil {")
-}
-gf.P("return err")
-gf.P("}")
+gf.P("if u, ok := any(item).(interface{ UnmarshalJSONSebuf([]byte, protojson.UnmarshalOptions) error }); ok {")
+gf.P("if err := u.UnmarshalJSONSebuf(itemRaw, opts); err != nil {")
+gf.P("return err")
+gf.P("}")
+gf.P("} else if err := opts.Unmarshal(itemRaw, item); err != nil {")
+gf.P("return err")
+gf.P("}")
@@
+gf.P("func (x *", msgName, ") UnmarshalJSON(data []byte) error {")
+gf.P("return x.UnmarshalJSONSebuf(data, protojson.UnmarshalOptions{})")
+gf.P("}")
+gf.P()
```

If a child message has its own `UnmarshalJSONSebuf`, the generated loop must call it with the same `opts` before falling back to `opts.Unmarshal`. Otherwise nested generated messages still lose the options.

It would be a good idea for a complete PR to apply this pattern to the root repeated, root map, and nested unwrap unmarshal paths in `internal/httpgen/unwrap.go`, regenerate the golden files, and add generated-client regression tests for all three paths.

MarkGHallak:
Fixes #204.

## What changed

- All three unwrap emitters in `internal/httpgen/unwrap.go` now emit `UnmarshalJSONSebuf(data []byte, opts protojson.UnmarshalOptions) error` as the real implementation plus a thin `UnmarshalJSON` wrapper that delegates with empty options. This mirrors what the same functions already do on the Marshal side and what `generateInt64UnmarshalJSON` does in `encoding.go`.
- Added `emitInlineUnmarshalChild` symmetric to `emitInlineMarshalChild` and used it at all six child unmarshal call sites, so each item dispatches through the child's own `UnmarshalJSONSebuf` when it has one and falls back to `opts.Unmarshal` otherwise. That made `hasEncodingMarshalJSON` dead, so it is deleted. `hasInt64NumberFields` stays because the encoding paths still use it.
- The bytes_encoding emitter had the same gap and is fixed the same way. Its generated `UnmarshalJSON` ended in a bare strict protojson call, so DiscardUnknown could never reach bytes encoded messages. It now emits `UnmarshalJSONSebuf` with the options threaded through plus the thin wrapper, mirroring the int64 emitter. The client generator's bytes emitter already had the twin, and the encoding files that real builds ship come from this plugin, so this closes the gap on the side real builds actually use.
- Regenerated the affected golden files, the three unwrap ones plus the bytes encoding one. The output matches the shape proposed in the issue.
- Behavior for existing callers is unchanged. The wrapper delegates with `protojson.UnmarshalOptions{}`, which is exactly what the package level `protojson.Unmarshal` does, and the strict default is pinned by the tests below.

## Enum types

Enums are the one place the invariant deliberately does not apply. Both generators emit enums with the plain `MarshalJSON`/`UnmarshalJSON` pair only, protojson options act on message fields and an enum scalar has none, so there is nothing for an options aware method to do. The pairing test exempts enums by their value receiver `MarshalJSON` rather than by file name, so the exemption is principled and the invariant stays tight for every message type in the files it scans.

## Tests

- `unwrap_test.go` gains `UnmarshalJSONSebuf` signature assertions for all four unwrap shapes, plus an Unmarshal side subtest in `TestCrossFileInt64EncodingUnwrap` asserting that no bare `protojson.Unmarshal(itemRaw, item)` or `json.Unmarshal(itemRaw, item)` bypass remains and that the inline forwarding is present.
- `unwrap_pairing_test.go` is new. It scans the checked in `*_unwrap.pb.go` and `*_encoding.pb.go` goldens and asserts every message type with a stdlib JSON method has its Sebuf twin and every Sebuf method has its stdlib wrapper. It reads the goldens directly so it needs no protoc, and it guards the unwrap and encoding emitters against this class of regression.
- `testdata/proto/unwrap.proto` gains a plain repeated message field on `MixedResponse`. One of the six child dispatch sites had no fixture anywhere, so reverting it left every generated file byte identical and no test failed. With the field in place the golden test catches that revert, and it pins the marshal side of the same branch as well.
- `examples/market-data-unwrap` is regenerated with the fixed plugins and now carries the executable proof, exercising every unwrap shape at runtime. `discard_unknown_test.go` drives the generated client against an httptest server whose hand written response contains a field the client's proto does not know, and asserts `WithMarketDataServiceDiscardUnknownFields(true)` decodes it while the strict default still rejects the same body. `unwrap_quick_check_test.go` keeps the issue's compile time quick check compiling and runs the two line repro directly on the checked in type. A new `unwrap_coverage.proto` plus `unwrap_shapes_test.go` add root repeated, root map with scalar and message values, and the combined map plus value unwrap shapes, each decoded at runtime with a strict and a DiscardUnknown pair. The last test there also proves the inline child dispatch takes the Sebuf branch at runtime, using a list whose elements are themselves root unwrap messages, since their bare array form only decodes through the element's own method. The lenient assertions fail against the current generator, which is the reason to check them in.
- `docs/json-protobuf-compatibility.md` and the example README now describe the options aware pair plus the wrappers.


internal/httpgen/unwrap_pairing_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	checked := 0
+	// Enum types are exempt. Protojson options act on message fields and an
+	// enum scalar has none, so enums carry only the plain pair. They are
+	// recognizable by their value-receiver MarshalJSON.
+	enumRe := regexp.MustCompile(`func \(x (\w+)\) MarshalJSON\(`)
+
+	for _, entry := range entries {
+		name := entry.Name()
+		if !strings.HasSuffix(name, "_unwrap.pb.go") && !strings.HasSuffix(name, "_encoding.pb.go") {
```
SebastienMelki: This filter scopes the invariant to exactly the two file kinds this PR already fixed, so the guard can only ever confirm what's already true. Widening it to every `*.pb.go` golden turns it into a real drift guard — the value-receiver enum exemption above already handles enums.

It would fail today on 8 types (`empty_behavior_empty_behavior.pb.go: Response`, `flatten_flatten.pb.go: DualFlatten/MixedFlatten/SimpleFlatten`, `nullable_nullable.pb.go: User`, `oneof_discriminator_oneof_discriminator.pb.go: FlattenedEvent/NestedEvent`, `timestamp_format_timestamp_format.pb.go: TimestampFormatTest`) — which is the honest signal. If those stay for a follow-up, an explicit skip list with a TODO beats a narrow suffix filter, because the narrow filter reads as "invariant holds everywhere" when it doesn't.

Minor: the regexes require the receiver to be named `x`. The `checked == 0` guard catches finding no *files*, but not finding no *types* — worth asserting a minimum type count too.

internal/httpgen/unwrap.go
```diff
@@ -733,13 +731,7 @@ func (g *Generator) generateScalarFieldUnmarshal(
 	gf.P(`if rawField, ok := raw["`, jsonName, `"]; ok {`)
 	if field.Message != nil {
 		gf.P("x.", fieldName, " = &", field.Message.GoIdent.GoName, "{}")
```
SebastienMelki: Pre-existing, but it's in the branch this PR touches: `field.Message.GoIdent.GoName` is passed as a plain string, so protogen can't qualify it. Every sibling site passes the `GoIdent` itself (or `gf.QualifiedGoIdent(...)`) and gets the import handled. A singular message field whose type comes from another Go package would emit a bare, unqualified type name here and fail to compile.

The mutation check says this site is covered by `unwrap_int64_encoding_unwrap.pb.go`, but only same-package. Worth a cross-package singular-message fixture alongside the cross-file `extra_bars` one.

internal/httpgen/unwrap.go
```diff
@@ -587,9 +591,9 @@ func (g *Generator) generateScalarFieldMarshal(
 func (g *Generator) generateUnwrapUnmarshalJSON(gf *protogen.GeneratedFile, containing *UnwrapContainingMessage) {
 	msgName := containing.Message.GoIdent.GoName
 
-	gf.P("// UnmarshalJSON implements json.Unmarshaler for ", msgName, ".")
+	gf.P("// UnmarshalJSONSebuf implements sebufUnmarshaler for ", msgName, ".")
```
SebastienMelki: Nit: `sebufUnmarshaler` is declared by clientgen in the *services* package, but these methods land in the models package, so the comment names a type that isn't in scope where it's emitted. The `sebufMarshaler` comments already do the same thing, so this is consistent with the existing convention — flagging only in case you'd rather the comment say "options-aware unmarshaler" and stop naming a type.

examples/market-data-unwrap/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/SebastienMelki/sebuf/examples/market-data-unwrap
 go 1.26.0
 
 require (
-	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.36.11-20260209202127-80ab13bee0bf.1
+	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.36.11-20260709200747-435963d16310.1
```
SebastienMelki: Unrelated dep bump rode in with the regen: this moves protovalidate to `…20260709200747` while `examples/market-data-unwrap/buf.lock` still pins commit `2a1774d888024a9b93ce7eb4b59f6a83`, and 10 other examples stay on `…20260209202127`. Either revert this pair of lines or bump every example (plus buf.lock) in its own PR — right now this example is the odd one out for no reason connected to the fix.

SebastienMelki:
## One change requested before approval

Everything else in my review is filed as follow-up work and is not this PR's job: #235 (five httpgen emitters still missing the twin), #236 (both plugins write identical filenames, so plugin order decides which implementation ships — the root cause of that class), #237 (examples never tested in CI; `make build` ignores `internal/` changes).

The one thing that should change here, because it is about this PR's own guard rather than new scope:

**`internal/httpgen/unwrap_pairing_test.go:37` — widen the file filter, with an explicit skip list.**

```go
if !strings.HasSuffix(name, "_unwrap.pb.go") && !strings.HasSuffix(name, "_encoding.pb.go") {
    continue
}
```

As written, the invariant is checked over exactly the two file kinds this PR fixes, so the guard can only confirm what is already true. Read cold, a test named `TestGoldenSebufMethodPairing` scanning `testdata/golden` implies the pairing holds for generated output generally. It does not: widened to every `*.pb.go` golden it fails on 8 types today (`empty_behavior_empty_behavior.pb.go: Response`, `flatten_flatten.pb.go: DualFlatten`/`MixedFlatten`/`SimpleFlatten`, `nullable_nullable.pb.go: User`, `oneof_discriminator_oneof_discriminator.pb.go: FlattenedEvent`/`NestedEvent`, `timestamp_format_timestamp_format.pb.go: TimestampFormatTest`).

Suggested shape — scan everything, name what is knowingly unfixed, and make the list self-deleting:

```go
// Emitters that still lack the twin, tracked in #235. Delete entries as they
// are fixed; the empty list is the goal state.
var pairingTODO = map[string]bool{
    "empty_behavior_empty_behavior.pb.go":         true,
    "flatten_flatten.pb.go":                       true,
    "nullable_nullable.pb.go":                     true,
    "oneof_discriminator_oneof_discriminator.pb.go": true,
    "timestamp_format_timestamp_format.pb.go":     true,
}

if !strings.HasSuffix(name, ".pb.go") || pairingTODO[name] {
    continue
}
```

That keeps this PR green, covers every golden the fix actually applies to, and makes #235 concrete instead of invisible. A `t.Log` naming the skipped files is a nice touch so the gap shows up in verbose output.

Two optional tightenings while you are in there: the regexes require the receiver to be named `x`, and the `checked == 0` guard catches finding no *files* but not finding no *types* — asserting a minimum type count closes both.

With that, this is an approve from me. Recap of what I verified: full suite green, `golangci-lint run ./internal/httpgen/...` clean (both `//nolint:dupl` removals are safe), example module tests pass, and mutation-testing each of the six dispatch sites (`unwrap.go:671, 713, 734, 958, 988, 1059`) fails a test every time — so the six-site claim holds and the `extra_bars` fixture is load-bearing.

examples/market-data-unwrap/go.mod
MarkGHallak: Correct me if I'm wrong but this version is required since #205. Without this bump the example doesn't build, so the tests this PR adds there can't run. I can revert this if you want the version to be uniform.

MarkGHallak:
Pushed 90f6217.

Pairing test now scans every `*.pb.go`, with a `pairingTODO` map naming the five files that still fail, a #235 reference, and a `t.Log` listing them. Regexes take any receiver name, and the `checked==0` guard is replaced by the minimum type count you suggested. The floor is 30 against the 39 the scan finds today. It sits just below the real count so it fails when the regexes stop matching, without breaking every time a fixture is added or removed.

`unwrap.go:733` passes the GoIdent instead of the bare GoName string. Added a cross-package fixture, `cross_pkg_meta.proto` and `cross_pkg_response.proto` in two Go packages, so the golden pins `x.Meta = &crosspkgmeta.Meta{}` plus the import. Reverting the line drops both and fails the compare.

The unwrap output no longer names `sebufMarshaler` or `sebufUnmarshaler`. `bytes_encoding.go`'s marshal comment keeps the old wording, since `TestGoGeneratorsProduceIdenticalBytesEncoding` compares that block against `clientgen `byte for byte. Happy to do `clientgen` in a follow-up PR so both sides match.

Reverted the protovalidate bump, so `go.mod` and `go.sum` are back to base and drop out of the diff. The example needs a go mod tidy before its tests run, same as the other examples. Happy to bump them all in a separate PR if you want them consistent.

Issue #213, opened 2026-07-21
enum_value / JSON-mapping marshalers: transitive nesting, generator composition, dual-plugin file collision, cross-package enums


SebastienMelki:
## Context

While fixing the server-side `enum_value` encoding bug (server emitted `RISK_LEVEL_LOW` where the OpenAPI docs / TS / Python clients used `low`), we deliberately scoped the fix to **direct enum fields on a message** and added a message-level marshaler (`*_enum_field_encoding.pb.go`) for the Go server and Go client. The following related issues were found but left out of scope. They should be tackled together since they share the same root cause: multiple annotation generators each want to own a message's single `MarshalJSON`/`MarshalJSONSebuf`.

## 1. Transitive nesting is not handled

A message whose **nested message-type field** references a custom-enum message is not patched — `protojson.Marshal` of the parent does not invoke the child's `MarshalJSONSebuf`, so the child's enum fields still serialize as raw proto names.

Example: `Parent { Child child = 1; }` where `Child` has a `RiskLevel risk = 1;` with `enum_value` mappings — `Parent` responses emit `RISK_LEVEL_LOW` for `child.risk`.

Fix direction: an enum wrapper pass analogous to `Int64WrapperContext` / `collectWrapperContexts` in `internal/httpgen/encoding.go`. Critically, it must **share the same wrapper method** with the int64 wrapper logic to avoid re-introducing a duplicate-`MarshalJSON` collision.

## 2. Field-level annotation generators do not compose (latent duplicate-method compile bug)

A Go type can only have one `MarshalJSON`/`MarshalJSONSebuf`/`UnmarshalJSON`. Today ~6 generators (unwrap, int64, bytes, empty_behavior, timestamp, nullable) each emit these per message and **assume sole ownership**. Coordination is only partial:

- `unwrap` ↔ `int64` have real bidirectional exclusion.
- `flatten` and `oneof` **fail-fast** when combined with the others.
- `int64` / `bytes` / `nullable` / `empty_behavior` / `timestamp_format` have **no pairwise guard** — a message with, e.g., an `int64_encoding=NUMBER` field **and** a `bytes_encoding=HEX` field emits `MarshalJSONSebuf` twice → **duplicate-method compile error**. The test suite never exercises these combos.

[...]

PR #214, opened 2026-07-21
fix(enum): apply enum_value custom strings on Go server & client JSON

SebastienMelki:
## Problem

Proto enums can carry `(sebuf.http.enum_value) = "low"` annotations to control JSON encoding. The **OpenAPI generator honors them** (emits `enum: [RISK_LEVEL_UNSPECIFIED, low, medium, high]`), and so do the TS/Python clients — but the generated **Go HTTP server emitted the raw proto names** (`RISK_LEVEL_LOW`), so the wire format contradicted the documented schema.

**Root cause:** both Go surfaces serialize at the *message* level with `protojson`:
- Server response: `protojson.Marshal(msg)`
- Server request: `protojson.Unmarshal(bodyBytes, msg)`
- Go client request: `protojson.Marshal(req)`

`protojson` serializes enums via proto reflection and **never invokes** the Go enum type's `MarshalJSON`/`UnmarshalJSON` (generated on the enum *type* in `*_enum_encoding.pb.go`). Those methods are only reached by stdlib `encoding/json`, so they were dead code on both Go paths.

## Fix

Add a **message-level** marshaler (`*_enum_field_encoding.pb.go`) for any message with a custom-enum field, mirroring the existing `int64_encoding`/`bytes_encoding` pattern. It reuses the existing `xToJSON`/`xFromJSON` lookup maps to rewrite enum fields between proto value names and custom strings, in **both directions** and across **all shapes**:

- singular, proto3 optional → `"low"`
- repeated → `["medium", "high"]`
- `map<_, enum>` values → `{"AAPL": "low", "TSLA": "high"}`

`MarshalJSONSebuf` emits the custom string; `UnmarshalJSONSebuf` accepts it (idempotent for clients that already send proto names or numbers). Mirrored in **clientgen** for Go-client request parity.

### Conflict guards
A Go type can own only one `MarshalJSON`. `enum_value` is registered in the existing fail-fast conflict web (`flatten`, `oneof`) plus a new `checkEnumMarshalJSONConflict`, so combining a custom enum with another JSON-mapping annotation errors clearly instead of emitting duplicate methods.

## Verification

```json
POST /api/v1/suggestion  {"symbol":"AAPL","requestedRisk":"low"}
-> {"riskLevel":"low","alternateRiskLevels":["medium","high"],
    "riskBySymbol":{"AAPL":"low","TSLA":"high"}, ...}
```

- New `examples/enum-encoding/` with an end-to-end test (`main_test.go`) proving the server emits custom strings and accepts them on input — no `RISK_LEVEL_` leakage.
- New golden files for httpgen + clientgen; `TestGoGeneratorsProduceIdenticalEnumFieldEncoding` confirms both Go generators are byte-identical.
- Fixture extended with a `map<string, Status>` field to cover the map branch; OpenAPI/TS goldens updated to match.
- Existing `TestGoGeneratorsProduceIdenticalEnumEncoding` stays green (`_enum_encoding.pb.go` untouched).
- Full suite + `go vet` pass.

## Note on the TimestampFormat golden

`TimestampFormatService.openapi.{yaml,json}` are regenerated here too. This golden was **already stale on current deps** (well-known `google.protobuf.Timestamp` comment text drifted from a recent protobuf bump) and its test fails on `main` without regenerating — unrelated to enums, folded in only because `UPDATE_GOLDEN` touched it.

## Out of scope (tracked in #213)

Transitive nesting (enum inside a nested message field), full field-patcher composition (+ the pre-existing int64/bytes latent duplicate-method bug), the dual-Go-plugin file collision (both plugins emit `_enum_encoding.pb.go`), and cross-package enum references.


SebastienMelki (commented):
Codex review feedback:

1. `UseProtoNames` regression: the enum field patcher always uses `field.Desc.JSONName()` keys, so `WithMarshalOptions(protojson.MarshalOptions{UseProtoNames: true})` emits snake_case fields that do not get patched. Repeated/map/custom enum fields can then leak raw proto enum names, and snake_case request bodies with custom enum strings also miss the rewrite before `protojson.Unmarshal`.

2. Cross-package enum gap: `customEnumFieldInfo` skips annotated enums unless `enum.GoIdent.GoImportPath == msg.GoIdent.GoImportPath`. Shared enums in another proto/Go package will silently keep proto-name JSON instead of the documented `enum_value` strings. If this is unsupported, generation should fail loudly; otherwise the generated lookup helpers need to be accessible/imported.

SebastienMelki:
Thanks for the review — both points addressed in 3c5f0ca:

**1. `UseProtoNames` regression** — fixed. The marshal and unmarshal patchers now iterate over both the camelCase JSON name and the snake_case proto name, e.g.:
```go
for _, k := range []string{"alternateRiskLevels", "alternate_risk_levels"} {
    v, ok := raw[k]
    ...
}
```
So repeated/map/singular enum fields are patched whether protojson emits camelCase or (under `UseProtoNames`) snake_case, in both directions. Added an end-to-end test in `examples/enum-encoding` (`TestEnumValueWithUseProtoNames`) that registers the server with `WithMarshalOptions(protojson.MarshalOptions{UseProtoNames: true})` and asserts `"risk_level":"low"` with no `RISK_LEVEL_` leakage.

**2. Cross-package enum gap** — now **fails loudly**. `validateEnumFieldEncoding` returns a generation error when a custom-`enum_value` field references an enum from a different Go package (the marshaler relies on that package's private `xToJSON`/`xFromJSON` maps). Chose fail-loud over silent proto-name output per your suggestion; full cross-package support (exported/qualified helpers) is tracked in #213.

SebastienMelki (commented):
Codex rereview: the previous feedback looks addressed.

- `UseProtoNames` now patches both camelCase and snake_case keys in the generated marshal/unmarshal paths.
- Cross-package annotated enums now fail generation loudly instead of silently emitting proto enum names.

Verified with:

```bash
GOCACHE=/private/tmp/go-cache GOPATH=/private/tmp/go go test ./internal/httpgen ./internal/clientgen
```

and from `examples/enum-encoding`:

```bash
GOCACHE=/private/tmp/go-cache GOPATH=/private/tmp/go go test .
```

PR #215, opened 2026-07-21
fix(enum): apply enum_value custom strings through nested messages (transitive)

SebastienMelki:
## Problem

v0.21.0 fixed `enum_value` encoding only for enums on the **directly-marshaled** message. In production (`sarwa-co/contracts`), the RPC returns a wrapper whose custom enums are **nested**:

```protobuf
message GetEasyOptionsResponse { repeated EasyOptionSuggestion data = 1; }   // marshaled here
message EasyOptionSuggestion  { OptionsContract contract = 1; RiskLevel risk_level = 3; }
message OptionsContract       { OptionType type = 2; }                        // enum 2 levels down
```

`protojson.Marshal(GetEasyOptionsResponse)` serializes the nested children itself and **never invokes** their generated `MarshalJSONSebuf`, so `risk_level` and `contract.type` leaked as `RISK_LEVEL_LOW` / `OPTION_TYPE_CALL`.

## Fix

The enum-field marshaler is now **unified and recursive**. Every message that either holds custom enums directly **or** nests (at any depth) a message that does gets one `MarshalJSONSebuf`/`UnmarshalJSONSebuf` that:

1. patches its own direct enum fields (as before), and
2. re-serializes nested singular/repeated message fields through the child's `MarshalJSONSebuf`, forwarding opts — mirroring the existing int64-wrapper delegation.

Custom strings now propagate through the whole tree, under default and `UseProtoNames` output. Delegation chain for the case above: `GetEasyOptionsResponse` → `EasyOptionSuggestion` (patches `risk_level`, re-serializes `contract`) → `OptionsContract` (patches `type`).

### Guardrails (fail loud, never silently wrong)
- Nested custom enums inside `map<_, message>` values → generation error (not yet supported).
- Cross-package direct enums → generation error (unchanged).
- Combining with another MarshalJSON feature, including int64 nested wrappers → fail fast.

## Verification

- **Validated against the real `sarwa-co/contracts` protos**: generates cleanly, and `GetEasyOptionsResponse`/`EasyOptionSuggestion` get the correct delegating marshalers.
- **`examples/enum-encoding` rebuilt** to mirror that structure and cover the full matrix — annotated vs unannotated × direct vs nested — asserted end-to-end at runtime:

```json
{"data":[{"contract":{"type":"call","sentiment":"SENTIMENT_BULLISH",...},"riskLevel":"low"}, ...],
 "overallRisk":"low","marketSentiment":"SENTIMENT_BULLISH"}
```
Annotated enums → custom strings at every depth; unannotated `Sentiment` → proto name kept.

- New three-level golden fixture (`enum_nested.proto`) with an unannotated enum verifying it is left untouched; `TestEnumFieldEncodingTransitiveNesting` + cross-generator identity checks; full suite + `go vet` + lint green. No openapi/ts/py golden churn (dedicated fixture).

Fixes the production issue behind #213 (item 1: transitive nesting).


SebastienMelki (commented):
Codex review feedback:

1. `UseProtoNames` still has a gap for nested message fields. In `internal/httpgen/enum_field_encoding.go` and `internal/clientgen/enum_field_encoding.go`, nested message marshal/unmarshal generation uses only `field.Desc.JSONName()`. Direct enum fields use `enumFieldJSONKeys` to cover both camelCase and proto snake_case, but nested fields do not.

With `protojson.MarshalOptions{UseProtoNames: true}`, `opts.Marshal(x)` emits the nested field under its proto name. The generated code then writes the patched child under the camelCase key, leaving the original snake_case field with raw proto enum names and adding a duplicate logical field. On unmarshal, snake_case nested fields are skipped, so custom nested enum strings are passed through to `protojson.Unmarshal` unchanged and can fail.

The example `UseProtoNames` test currently uses nested field names like `data`, `contract`, and `type`, where JSON/proto names are identical, so it misses this case. Nested fields should use the same key-list approach as direct enum fields and update whichever key is present.

Verified during review:

```bash
GOCACHE=/private/tmp/go-cache GOPATH=/private/tmp/go go test ./internal/httpgen ./internal/clientgen
```

and from `examples/enum-encoding`:

```bash
GOCACHE=/private/tmp/go-cache GOPATH=/private/tmp/go go test .
```

SebastienMelki:
Good catch — fixed in 8797b30.

The nested message marshal/unmarshal now use the same both-keys approach as direct enum fields (`enumFieldJSONKeys`), patching whichever key protojson emitted:

```go
childData, childErr := m.MarshalJSONSebuf(opts)
...
for _, k := range []string{"optionsContract", "options_contract"} {
    if _, ok := raw[k]; ok { raw[k] = childData }
}
```

Unmarshal likewise loops over both keys and updates whichever the request used.

To make this regression-proof (not just pass by coincidence of single-word names), I renamed the nested container fields to multi-word ones — `item_list`/`lead_item`/`item_group` in the golden fixture, and `option_suggestions`/`options_contract` in `examples/enum-encoding`. The golden consistency test now asserts `[]string{"itemList", "item_list"}` etc., and the example's `UseProtoNames` test asserts the custom strings appear under the snake_case keys **and** that there are no duplicate camelCase keys. Verified end-to-end:

```
UseProtoNames: {"option_suggestions":[{"options_contract":{..."type":"call"},"risk_level":"low"}],"overall_risk":"low"}
```

Note: the pre-existing int64 wrapper (`internal/httpgen/encoding.go`) has the same JSONName-only limitation I mirrored from; I left it untouched here to keep the PR focused — worth a follow-up.

SebastienMelki (commented):
Codex rereview: no blocking findings. The previous nested `UseProtoNames` issue looks addressed: nested message fields now patch both JSON key spellings via `enumFieldJSONKeys`, and the generated output updates whichever key is present without adding duplicate camelCase fields.

Looks safe to merge from this review.

Verified with:

```bash
GOCACHE=/private/tmp/go-cache GOPATH=/private/tmp/go go test ./internal/httpgen ./internal/clientgen
```

and from `examples/enum-encoding`:

```bash
GOCACHE=/private/tmp/go-cache GOPATH=/private/tmp/go go test .
```

Issue #165, opened 2026-05-07
[FEATURE] Support nested message fields in query parameters via field-path keys


AliMsayleb:
**Is your feature request related to a problem? Please describe.**

When a request message uses a nested message field annotated with `(sebuf.http.query)`, the generated HTTP handler returns a validation error at request time:

```json
{
  "violations": [
    {
      "field": "page",
      "description": "invalid value for query parameter page: unsupported field type: message"
    }
  ]
}
```

This makes it impossible to reuse a common request type (like a shared `PageRequest`) across GET endpoints, forcing every list request to inline the same scalar fields (`page_size`, `page_token`, ...) and creating drift over time.

**Describe the solution you'd like**

Bind nested message fields from **field-path query keys**, the convention adopted by gRPC-Gateway and Google's HTTP transcoding:

```
GET /items?page.page_size=20&page.page_token=abc
```

This requires:
- `protoc-gen-go-http` to emit one `QueryParamConfig` per leaf scalar inside the message tree, with a dotted query name and a multi-segment field path.
- `bindQueryParams` (in conjunction with `convertStringToFieldValue`) to walk message reflection along the field path before assigning the value.
- Recursion for arbitrary nesting depth (`?a.b.c=...`).
- Repeated leaf fields keep the standard repeated-query convention (`?page.tags=a&page.tags=b`).

**Describe alternatives you've considered**

1. **Inline scalars on every request** — copy `page_size`/`page_token` into every `List*Request`. Works today but creates drift across services and prevents centralizing validation rules.
2. **JSON-encoded blob in a single query param** — e.g. `?page={"page_size":20}`. Works mechanically but ugly URLs, no OpenAPI affordance, escaping headaches.
3. **Switch list endpoints to POST** — typed nested messages bind cleanly via the JSON body, but loses GET semantics (cacheability, idempotency, REST conventions).
4. **Adopt gRPC-Gateway** — solves the problem but is a much larger framework swap.

**Use Cases**

1. **Reusable pagination type** — a `PageRequest` (`page_size`, `page_token`) defined once in a `core/` proto and embedded in every list request. Single source of truth for pagination semantics and validation.
2. **Reusable sort/filter parameter groups** — e.g. `SortOptions { string order_by; SortDirection direction; }` reused across endpoints: `?sort.order_by=created_at&sort.direction=DESC`.
3. **Common filter envelopes** — date ranges, geo bounding boxes, etc., factored into shared messages so search endpoints share a vocabulary.

[...]

Issue #186, opened 2026-06-09
codegen: repeated enum query params generate invalid Go comparison


SebastienMelki:
## Bug

The Go client codegen for **repeated enum** query parameters emits `req.Field != 0`, but the field type is a slice (`[]EnumType`), so this doesn't compile.

### Affected generated code

```go
// sdk/search/v1/service_client.pb.go:189
if req.Types != 0 {  // Types is []sharedv1.ContentType — slice, not scalar

// sdk/search/v1/service_client.pb.go:347
if req.ContentTypes != 0 {

// sdk/soundtrack/v1/service_client.pb.go:268
if req.ContentTypes != 0 {
```

### Expected

For repeated fields (including repeated enums), the zero-value check should be:

```go
if len(req.ContentTypes) != 0 {
```

### Reproduction

Any proto service with a repeated enum query parameter, e.g.:

```proto
rpc Search(SearchRequest) returns (SearchResponse) {
  option (google.api.http) = { get: "/v1/search" };
}

message SearchRequest {
  repeated ContentType types = 1;
}
```

Generates a client with `req.Types != 0` which fails to compile.

### Context

Introduced in sebuf v0.16.0 (the enum query-param fix from #185). The scalar enum case works correctly — this only affects **repeated** enum fields.

PR #218, opened 2026-07-27
fix(codegen): reject non-scalar query and path parameters at generation time

### LINKED ISSUE #216: codegen: singular message-kind query params break go-client (compile), py/ts-client (wrong value), go-http (400), and openapiv3 (mismatched contract)
## Bug

A **singular message-kind** field annotated with `(sebuf.http.query)` is mishandled by every generator that consumes `annotations.GetQueryParams`. The Go client doesn't compile; the Python client silently sends a garbage value; the Go HTTP server rejects the request at runtime; and OpenAPI documents a contract none of them implement.

The common cause: `annotations.GetQueryParams` happily returns message-kind fields (`internal/annotations/query.go:61` sets `FieldKind: field.Desc.Kind().String()` → `"message"`), and every consumer's kind switch falls through to a scalar/string `default:` branch. Nothing validates or handles the message case.

This is the same class as #186 (repeated enum → `req.Field != 0`), one type-kind over. Related to #165, but distinct: #165 asks for *nested* messages to bind via dotted field-path keys and is scoped to `gen/go-http`; this report is about a **single-scalar wrapper message** (the typed-ID pattern) used directly as a query param, and the headline symptom is a **Go client compile break**, not a 400.

### Reproduction

```protobuf
// contracts/core/v1/identifiers.proto
message UserClientID {
  string value = 1 [(buf.validate.field).string.uuid = true];
}

// contracts/authorization/v1/list_client_restrictions.proto
message ListClientRestrictionsRequest {
  contracts.core.v1.UserClientID client_id = 1 [
    (buf.validate.field).required = true,
    (sebuf.http.query) = {
      name: "clientId"
      required: true
    }
  ];
}

service AuthorizationService {
  rpc ListClientRestrictions(ListClientRestrictionsRequest) returns (ListClientRestrictionsResponse) {
    option (sebuf.http.config) = {path: "/clients/restrictions", method: HTTP_METHOD_GET};
  }
}
```

Observed with sebuf **v0.21.1**.

### Symptom 1 — `protoc-gen-go-client`: generated code does not compile

```go
// api/contracts/authorization/v1/service_client.pb.go:304
if req.ClientId != "" {   // req.ClientId is *corev1.UserClientID
    queryParams.Set("clientId", fmt.Sprint(req.ClientId))
}
```

```
contracts/authorization/v1/service_client.pb.go:304:21: invalid operation: req.ClientId != "" (mismatched types *corev1.UserClientID and untyped string)
contracts/authorization/v1/service_client.pb.go:377:21: invalid operation: req.ClientId != "" (mismatched types *corev1.UserClientID and untyped string)
```

Source: `internal/clientgen/generator.go:845` emits `if req.X != <getZeroValue(qp)>`, and `getZeroValue` (`:1034`) has no `message` case, so the `default:` at `:1048` returns `""`.

Even if the guard compiled, `fmt.Sprint(req.ClientId)` on a proto message would emit a Go struct dump, not the UUID.

### Symptom 2 — `protoc-gen-py-client`: silently sends the wrong value

```python
# sdk/python/contracts/authorization/v1/service_client.py:571
if req.client_id is not None and req.client_id != 0:
    query_pairs.append(("clientId", str(req.client_id)))
```

Two problems: `req.client_id != 0` compares a dataclass to `0` (always true, harmless but meaningless), and `str(req.client_id)` yields the dataclass repr — `UserClientID(value='1ebfbcfd-…')` — which is URL-encoded and sent as the `clientId` value. This one is worse than the Go break because it fails silently at runtime rather than at build time.

Source: `internal/pyclientgen/client.go:197-199` — the `default:` branch, whose `//nolint:exhaustive` comment at `:189` explicitly lumps message kinds in with numerics: *"default covers all numeric/message kinds with a single expression"*.

### Symptom 3 — `protoc-gen-go-http`: server can't parse it either

`convertStringToFieldValue` (`internal/httpgen/generator.go:679`) switches on `field.Kind()` with no `MessageKind` case, so `default:` (`:739-740`) returns `unsupported field type: message`, surfaced as a 400:

```json
{"violations":[{"field":"client_id","description":"invalid value for query parameter clientId: unsupported field type: message"}]}
```

Same runtime error reported in #165.

### Symptom 4 — `protoc-gen-openapiv3`: documents a contract nothing implements

```yaml
- name: clientId
  in: query
  required: true
  schema:
    type: string     # flattened to a string...
```

…while the same field in a request *body* is `$ref: '#/components/schemas/UserClientID'` (an object). The `type: string` isn't deliberate unwrapping — it's the `default:` fallback in `createScalarFieldSchema` (`internal/openapiv3/generator.go:1032-1033`) treating unknown kinds as strings. It happens to describe the most sensible wire format, which no generator actually produces or accepts.

### Symptom 5 — `protoc-gen-ts-client`: same fallthrough

`tscommon.TSZeroCheck` (`internal/tscommon/types.go:96`) has no message case; `default:` produces a truthy check. TS is loose enough to compile, but the serialized value has the same problem as Python.

## Expected

For a wrapper message holding exactly one scalar field — the typed-ID pattern (`UserClientID{ string value = 1 }`), which is pervasive in our contracts — flatten to that scalar on the wire. That matches what OpenAPI already documents (`type: string`) and keeps URLs clean (`?clientId=1ebfbcfd-…`):

```go
// go-client
if req.ClientId != nil && req.ClientId.GetValue() != "" {
    queryParams.Set("clientId", fmt.Sprint(req.ClientId.GetValue()))
}
```

```python
# py-client
if req.client_id is not None and req.client_id.value:
    query_pairs.append(("clientId", str(req.client_id.value)))
```

```go
// go-http: convertStringToFieldValue needs a MessageKind case that, for a
// single-scalar wrapper, allocates the message and sets its sole field.
```

Multi-field / nested messages are the #165 design (dotted field-path keys) and could stay unsupported here — but they should fail at **generation** time with a clear error rather than emitting code that doesn't compile (Go), silently corrupts the value (Python/TS), or 400s at runtime (server).

### Suggested shape of the fix

1. Add a shared helper in `internal/annotations` — e.g. `func (qp QueryParam) WrapperScalarField() *protogen.Field` — returning the sole scalar field when the param is a message with exactly one non-repeated scalar field, else nil. All five generators consume the one helper so they can't drift again.
2. `clientgen`, `pyclientgen`, `tsclientgen`: when the helper returns a field, emit nil-check + `.GetValue()`-style access and encode the inner scalar's kind.
3. `httpgen`: add a `MessageKind` case to `convertStringToFieldValue` that populates the wrapper's sole field.
4. `openapiv3`: derive the query param schema from the inner scalar (including `format: uuid` and examples), so the docs match by construction instead of by accident.
5. Reject message-kind query params the helper can't flatten with a generation-time error naming the field, and add golden-file coverage for both the wrapper and the rejected case — mirroring how #186 was closed.

SebastienMelki:
Closes #216.

## The bug

A singular message-kind field annotated with `(sebuf.http.query)` — the typed-ID wrapper pattern — was mishandled by **every** generator. `annotations.GetQueryParams` returns it with `FieldKind: "message"` (`internal/annotations/query.go:61`) and each consumer's kind switch falls through to a string-shaped `default:`.

I verified all five reported symptoms at the exact lines cited in the issue:

| Generator | Symptom |
|---|---|
| `go-client` | emits `if req.X != ""` on a `*Msg` → **does not compile** |
| `py-client` | emits `req.x != 0` + `str(req.x)` → sends the **dataclass repr** |
| `ts-client` | emits `String(req.x)` → sends `"[object Object]"` |
| `ts-server` | assigns a `string` into a message-typed property |
| `go-http` | runtime **400** `unsupported field type: message` |
| `openapiv3` | documents `type: string` — a contract nothing implements |

## Why rejection instead of flattening

The issue proposes flattening single-scalar wrapper messages to their inner scalar. This PR deliberately does not do that:

1. **Rejection is reversible.** Flattening invents a wire-format rule (message *shape* determines encoding) that could never be removed once shipped. Rejection can be relaxed into flattening later; the reverse isn't true.
2. **It matches a decision this repo already made.** `isPathParamCompatible` already rejected `MessageKind`/`GroupKind`/`BytesKind` for path params. Rejecting for query params aligns the two — the flattening proposal would have *reversed* that stance for both.
3. **It breaks no working code.** All six consumers were already broken. One clear generation-time error is strictly better than every current outcome.

**Accepted cost:** the typed-ID + GET pattern is no longer expressible. The workaround keeps validation and lets the wrapper type stay in use everywhere else:

```protobuf
string client_id = 1 [
  (buf.validate.field).string.uuid = true,
  (sebuf.http.query) = {name: "clientId", required: true}
];
```

## Scope

- Rejects **message, group, and bytes**. `bytes` hits the identical `default:` branch in `convertStringToFieldValue` — same bug, one type-kind over. Map fields report `MessageKind` and are covered by the same predicate.
- Covers **both query and path** params, wired into **all six** generators.

## Two structural problems the issue missed

**Validation only fired under `protoc-gen-go-http`.** `ValidateService` had exactly one call site. Anyone generating only the TS or Python client got *no* validation and still received broken output. The new check is wired into all six entry points.

**A shared helper does not prevent drift — wiring does.** `annotations.ValidateTimestampFormatAnnotation` is already a shared helper, called by only 2 of 6 generators. The testdata has drifted too: `internal/pyclientgen/testdata/proto/query_params.proto` is already out of sync with its five byte-identical siblings. So this PR adds a cross-generator consistency test rather than relying on the shared function alone.

## Implementation

`internal/annotations/url_params.go` is the single source of truth: `IsURLParamKindCompatible` (kind predicate), `URLParamValidationError` (one shared message formatter), `ValidateFileURLParams` (per-file entry point).

httpgen routes it through its existing aggregating `ValidateMethodConfig`; the other five call `ValidateFileURLParams` before emitting anything, so an invalid proto never produces partial output. httpgen's local `isPathParamCompatible` and `findFieldByProtoName` are deleted in favour of the shared helpers, and its test that *mirrored* the predicate's switch now calls the real function.

Errors name the field, message, parameter, and offending type:

> `AuthorizationService.ListClientRestrictions: field 'client_id' on message 'ListClientRestrictionsRequest' is annotated with (sebuf.http.query) as parameter 'clientId', but has unsupported type 'message (core.v1.UserClientID)'. Query parameters must be scalar types (…) or repeated scalars. Replace it with a scalar field, or move it into the request body by using POST/PUT/PATCH.`

## Tests

`internal/urlparamtest` drives all six generators over one shared set of fixtures (singular message, repeated message, map, bytes, message path param, plus a valid scalar positive control) — 42 subtests asserting they reject *and* accept identically.

The fixtures live in one place rather than six copies: the per-generator-copy convention exists so golden files stay independent, these produce no goldens, and six copies is exactly how `query_params.proto` drifted.

`internal/annotations/url_params_test.go` covers the new validator to 100% using synthesized descriptors (no protoc needed).

## Verification

- `go test ./...` green; `scripts/run_tests.sh` — **12/12 packages pass the coverage gate**.
- **Zero golden churn** — regenerated every golden across all six generators; `git diff --exit-code -- '*testdata/golden*'` clean.
- End-to-end through real `protoc`: full 6-plugin × 6-fixture matrix, all reject/accept correctly.
- Lint back to the exact pre-existing baseline.

## Release note

**Minor version bump, not a patch.** Query annotations are not gated by HTTP method in httpgen (`generator.go:430`, `:1823`) or openapiv3 (`generator.go:894`) — only clientgen limits them to GET/DELETE. So a `(sebuf.http.query)` sitting inertly on a message field in a POST request compiled before and will now fail generation. A nonsense annotation, but a real behavioural break.

## Follow-ups (not in this PR)

- **Makefile stale-binary bug**: the pattern rule `$(BIN_DIR)/%: $(CMD_DIR)/%/*.go` depends only on `cmd/`, not `internal/`, so `make build` silently keeps stale plugin binaries when only generator internals change. This misled my first end-to-end run.
- **Typed-ID wrapper flattening** as an explicit opt-in feature — the reporter's original ask. Related to #165 (nested messages via dotted field-path keys).
- **#219** — `enum_value` custom strings are ignored in URL parameter encoding, so the TS client
  sends `americas` where the Go client/server use `REGION_AMERICAS`, and the Go server rejects it.
  Now filed with a full scope analysis. (My original note here also listed `int64_encoding` and
  `bytes_encoding` — those turned out **not** to be bugs: int64 encoding is meaningless in a URL,
  where values are always strings, and bytes/Timestamp are rejected outright by this PR.)
- ~~`isPathParamCompatible` checked `Kind()` but not `IsList()`~~ — **fixed in this PR** (`60e39e6`), after review.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -547,6 +547,28 @@ All custom annotations live in `proto/sebuf/http/annotations.proto`:
 | 50019 | flatten | FieldOptions | Nested message flattening |
 | 50020 | flatten_prefix | FieldOptions | Prefix for flattened fields |
 
+### URL Parameter Types Must Be Scalar
+
+Query (`sebuf.http.query`) and path parameters accept only scalar kinds — string,
+bool, the integer families, float/double, and enum — plus repeated scalars for query
+params. `message`, `group`, `bytes`, and `map` have no canonical URL encoding and are
+**rejected at generation time by all six generators**.
+
+The single source of truth is `internal/annotations/url_params.go`:
+`IsURLParamKindCompatible` (the kind predicate) and `ValidateFileURLParams` (the
+per-file entry point every generator calls). httpgen routes it through its own
+aggregating `ValidateMethodConfig`; the other five call `ValidateFileURLParams`
+directly before emitting anything.
+
+A shared helper alone does not keep the plugins aligned — `ValidateTimestampFormatAnnotation`
+is shared too and only two of six ever wired it in. `internal/urlparamtest` is the
+anti-drift guard: it drives all six generators over the same fixtures and asserts they
+reject and accept identically. Add a generator there when you add a generator.
+
+Notably, this makes the **typed-ID pattern** (`UserClientID { string value = 1 }`) invalid
+as a query/path param; use a scalar field with `(buf.validate.field).string.uuid` instead.
+See [docs/http-generation.md](docs/http-generation.md#unsupported-parameter-types) and issue #216.
+
 ## Development Commands
 
 ### Testing
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -550,13 +550,20 @@ All custom annotations live in `proto/sebuf/http/annotations.proto`:
 ### URL Parameter Types Must Be Scalar
 
 Query (`sebuf.http.query`) and path parameters accept only scalar kinds — string,
-bool, the integer families, float/double, and enum — plus repeated scalars for query
-params. `message`, `group`, `bytes`, and `map` have no canonical URL encoding and are
-**rejected at generation time by all six generators**.
+bool, the integer families, float/double, and enum. `message`, `group`, `bytes`, and
+`map` have no canonical URL encoding and are **rejected at generation time by all six
+generators**.
+
+Cardinality differs by location: repeated scalars are valid query params
+(`?ids=a&ids=b`) but never valid path variables, since a path variable matches exactly
+one URL segment. Accepting a repeated path variable makes the generated Go server
+*panic* at request time — `bindPathParams` calls `reflectMsg.Set(field, scalarValue)`,
+which protoreflect rejects on a list field.
 
 The single source of truth is `internal/annotations/url_params.go`:
-`IsURLParamKindCompatible` (the kind predicate) and `ValidateFileURLParams` (the
-per-file entry point every generator calls). httpgen routes it through its own
+`IsURLParamKindCompatible` (the kind predicate), `ValidatePathParamField` (kind +
+cardinality, for path variables), and `ValidateFileURLParams` (the per-file entry point
+every generator calls). httpgen routes it through its own
 aggregating `ValidateMethodConfig`; the other five call `ValidateFileURLParams`
 directly before emitting anything.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -583,13 +583,20 @@ All custom annotations live in `proto/sebuf/http/annotations.proto`:
 ### URL Parameter Types Must Be Scalar
 
 Query (`sebuf.http.query`) and path parameters accept only scalar kinds — string,
-bool, the integer families, float/double, and enum — plus repeated scalars for query
-params. `message`, `group`, `bytes`, and `map` have no canonical URL encoding and are
-**rejected at generation time by all six generators**.
+bool, the integer families, float/double, and enum. `message`, `group`, `bytes`, and
+`map` have no canonical URL encoding and are **rejected at generation time by all six
+generators**.
+
+Cardinality differs by location: repeated scalars are valid query params
+(`?ids=a&ids=b`) but never valid path variables, since a path variable matches exactly
+one URL segment. Accepting a repeated path variable makes the generated Go server
+*panic* at request time — `bindPathParams` calls `reflectMsg.Set(field, scalarValue)`,
+which protoreflect rejects on a list field.
 
 The single source of truth is `internal/annotations/url_params.go`:
-`IsURLParamKindCompatible` (the kind predicate) and `ValidateFileURLParams` (the
-per-file entry point every generator calls). httpgen routes it through its own
+`IsURLParamKindCompatible` (the kind predicate), `ValidatePathParamField` (kind +
+cardinality, for path variables), and `ValidateFileURLParams` (the per-file entry point
+every generator calls). httpgen routes it through its own
 aggregating `ValidateMethodConfig`; the other five call `ValidateFileURLParams`
 directly before emitting anything.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -580,6 +580,28 @@ All custom annotations live in `proto/sebuf/http/annotations.proto`:
 | 50019 | flatten | FieldOptions | Nested message flattening |
 | 50020 | flatten_prefix | FieldOptions | Prefix for flattened fields |
 
+### URL Parameter Types Must Be Scalar
+
+Query (`sebuf.http.query`) and path parameters accept only scalar kinds — string,
+bool, the integer families, float/double, and enum — plus repeated scalars for query
+params. `message`, `group`, `bytes`, and `map` have no canonical URL encoding and are
+**rejected at generation time by all six generators**.
+
+The single source of truth is `internal/annotations/url_params.go`:
+`IsURLParamKindCompatible` (the kind predicate) and `ValidateFileURLParams` (the
+per-file entry point every generator calls). httpgen routes it through its own
+aggregating `ValidateMethodConfig`; the other five call `ValidateFileURLParams`
+directly before emitting anything.
+
+A shared helper alone does not keep the plugins aligned — `ValidateTimestampFormatAnnotation`
+is shared too and only two of six ever wired it in. `internal/urlparamtest` is the
+anti-drift guard: it drives all six generators over the same fixtures and asserts they
+reject and accept identically. Add a generator there when you add a generator.
+
+Notably, this makes the **typed-ID pattern** (`UserClientID { string value = 1 }`) invalid
+as a query/path param; use a scalar field with `(buf.validate.field).string.uuid` instead.
+See [docs/http-generation.md](docs/http-generation.md#unsupported-parameter-types) and issue #216.
+
 ## Development Commands
 
 ### Testing
--- a/cmd/protoc-gen-openapiv3/main.go
+++ b/cmd/protoc-gen-openapiv3/main.go
@@ -109,6 +109,14 @@ func createPlugin(req *pluginpb.CodeGeneratorRequest) *protogen.Plugin {
 }
 
 func generateOpenAPIFiles(plugin *protogen.Plugin, format openapiv3.OutputFormat, bundle bundleConfig) {
+	// Reject path/query params whose types cannot be represented in a URL before
+	// emitting anything, so an invalid proto fails cleanly instead of documenting a
+	// contract no generator implements.
+	if err := openapiv3.ValidateFiles(plugin); err != nil {
+		plugin.Error(err)
+		return
+	}
+
 	// Per-service output (default behaviour; suppressed when bundle_only=true).
 	if !bundle.enabled || !bundle.onlyBundle {
 		for _, file := range plugin.Files {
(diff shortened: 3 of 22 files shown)


SebastienMelki:
Review finding:

- `internal/annotations/url_params.go`: `IsURLParamCompatible` only checks `field.Desc.Kind()`, so a repeated scalar field like `repeated string ids = 1` bound to `/items/{ids}` is still accepted. The PR docs and error wording say repeated fields are only supported for query parameters. This leaves broken generated behavior: the Go server converts the path segment to a scalar and then calls `reflectMsg.Set(field, convertedValue)` on a list field, and the other generators/OpenAPI encode or document incompatible shapes. Please reject `field.Desc.IsList()` for path params and add a cross-generator negative fixture for repeated-scalar path params.

I also ran the targeted package tests:

`go test ./internal/annotations ./internal/urlparamtest ./internal/httpgen ./internal/clientgen ./internal/tsclientgen ./internal/tsservergen ./internal/pyclientgen ./internal/openapiv3`

They pass with elevated permissions. The initial sandboxed run only failed due DNS access for `go mod tidy` and local listener binding in `httptest`.

SebastienMelki:
> Posted by Claude (via Claude Code) on behalf of @SebastienMelki.

## Review finding addressed in `60e39e6`

Good catch, and it was worse than described — I verified the runtime impact before fixing.

`reflectMsg.Set(field, scalarValue)` on a list field doesn't just produce an incompatible shape, it **panics**:

```
field "dependency" IsList=true kind=string
PANIC as predicted: type mismatch: cannot convert string to list
```

So a `repeated` path variable crashes the generated Go handler at request time rather than returning a 400. You were also right that this PR made the gap worse rather than merely leaving it: the docs and error wording I added explicitly state repeated is query-only, while the code still accepted it.

### Fix

Added `ValidatePathParamField(field, paramName, messageName)` as the single source of truth for path-variable binding — kind **and** cardinality together. Both `annotations.ValidatePathParams` and httpgen's aggregating `validatePathVariables` call it, so the two can't disagree. This removes the last place where the path rule was expressed twice.

One deliberate ordering choice: **kind is reported ahead of cardinality**. For a `repeated SomeMessage` path variable both are wrong, but dropping the repeated label alone still wouldn't make it bindable, so "must be scalar types" is the more actionable message. There's a unit case pinning that precedence.

`URLParamTypeName` now prefixes `repeated `, so errors read `'repeated string'` instead of `'string'`:

```
ItemsService.GetItems: path variable '{ids}' is bound to field 'ids' on message
'GetItemsRequest' of type 'repeated string', but a path variable matches a single URL
segment and cannot be repeated. Remove the repeated label, or bind the field as a query
parameter with (sebuf.http.query) instead.
```

### Repeated query params are unaffected

To be explicit, since it's the obvious regression risk:

| | Query param | Path variable |
|---|---|---|
| `repeated string` | ✅ `?tags=a&tags=b` | ❌ rejected (new) |
| `repeated <enum>` | ✅ | ❌ rejected |
| `repeated <message>` | ❌ | ❌ |

The repeated check lives only in `ValidatePathParamField`; the query path never calls it. The "must accept" fixture deliberately carries `repeated string tags` and `repeated Region regions` as query params, and the Go client golden still emits `queryParams.Add("regions", …)` — the #186 behaviour — with **zero golden diff against `main`**.

### Tests

- `invalid_repeated_scalar_path_param.proto` — new cross-generator negative fixture, asserted across all six generators (as requested).
- Unit cases for repeated-scalar and repeated-message path variants.
- Verified **no existing proto in the repo** binds a repeated field to a path variable, so nothing that previously generated now fails.
- Full suite green, no golden churn, lint back at the pre-existing baseline, validator still at 100% coverage.

### On the sandboxed run

That matches what I saw: the golden and consistency tests shell out to `protoc` and `t.Skip` when it's absent, and the httpgen suite binds local listeners via `httptest`. Worth noting the skip is silent — a sandboxed run reports `ok` while having exercised very little. Related: `make build`'s pattern rule `$(BIN_DIR)/%: $(CMD_DIR)/%/*.go` depends only on `cmd/`, not `internal/`, so it silently keeps stale plugin binaries when generator internals change. That misled my first end-to-end run — 5 of 6 plugins appeared to accept an invalid proto until I did `rm -rf bin && make build`. Both are pre-existing and worth separate issues.

SebastienMelki:
Follow-up review on latest head `60e39e6`:

The previous finding is addressed. The update adds `ValidatePathParamField`, routes httpgen through it, rejects `field.Desc.IsList()` for path variables, and adds `invalid_repeated_scalar_path_param.proto` to the cross-generator consistency matrix.

I re-ran the targeted verification:

`go test ./internal/annotations ./internal/urlparamtest ./internal/httpgen ./internal/clientgen ./internal/tsclientgen ./internal/tsservergen ./internal/pyclientgen ./internal/openapiv3`

All passed.

I did not find any new blocking issues. The only minor cleanup I noticed is that the PR body still lists “`repeated string` in a path still slips through” under follow-ups, which is now stale after this update.

From my side, this is good to go once that stale PR-body note is removed or updated.

PR #220, opened 2026-07-27
fix(codegen): generate transitive int64 NUMBER marshalers across files and at any depth

### LINKED ISSUE #217: httpgen: transitive int64_encoding=NUMBER wrapper marshaler not generated when nested message is in another file
## Summary

The transitive (wrapper) `MarshalJSONSebuf` for `int64_encoding = INT64_ENCODING_NUMBER` is only generated when the annotated message is declared in the same `.proto` file as the message containing it. When the annotated message comes from an imported file, no wrapper is generated and `protojson` emits the int64 as a quoted string.

## Reproduction

`internal/httpgen/testdata/proto/int64_nested_encoding.proto` declares `SensorReading` and `GetSensorReadingResponse` in one file, and the transitive marshaler is generated correctly. Split the same case across two files:

`reading.proto`:

```protobuf
syntax = "proto3";

package example.v1;

import "sebuf/http/annotations.proto";

message SensorReading {
  int64 timestamp_ms = 1 [(sebuf.http.int64_encoding) = INT64_ENCODING_NUMBER];
}
```

`response.proto`:

```protobuf
syntax = "proto3";

package example.v1;

import "reading.proto";

message GetSensorReadingResponse {
  SensorReading reading = 1;
}
```

**Expected:** `response_encoding.pb.go` containing `GetSensorReadingResponse.MarshalJSONSebuf`, forwarding opts to `reading`.

**Actual:** no file is generated. Output is:

```json
{"reading":{"timestampMs":"1715000000000"}}
```

instead of:

```json
{"reading":{"timestampMs":1715000000000}}
```

Same behaviour with `repeated SensorReading readings = 1`.

## Root cause

`internal/httpgen/encoding.go:170-179` builds `directMsgNames` from `collectInt64EncodingContext(file)` — messages declared in the file currently being generated only:

```go
contexts := collectInt64EncodingContext(file)
directMsgNames := /* built from contexts */
wrapperContexts := collectWrapperContexts(file, directMsgNames, unwrapMsgNames)
```

The qualifying test for a nested field is at `:131`:

```go
directMsgNames[string(field.Message.Desc.FullName())]
```

Imported types are never present in that set, so cross-file nested fields never qualify and no wrapper is emitted.

## Also affects depth > 1

Wrapper messages are never added to `directMsgNames`, so nesting deeper than one level is missed even within a single file. Given `A{B}`, `B{C}`, `C{int64 NUMBER}`, a wrapper is generated for `B` but not for `A`.

## Suggested fix

Replace the `:131` set lookup with a recursive predicate on `field.Message` — does this message transitively contain an int64 NUMBER field? That is file-independent and fixes depth > 1 in the same change. The generated wrapper only *calls* the child's exported `MarshalJSONSebuf`, so cross-file and cross-package both work.

Note that `internal/httpgen/cross_int64_consistency_test.go` covers this class of bug for the unwrap generator's map-element path only; `collectWrapperContexts` never received the equivalent fix.

Observed with v0.21.1; code path unchanged on `main` @ `ad6a1b5`.

SebastienMelki:
Fixes #217.

## Problem

The Go generators emitted a transitive `MarshalJSONSebuf` for a message nesting an `int64_encoding=NUMBER` type **only when both were declared in the same `.proto` file, and only at exactly one level of nesting**. Otherwise no wrapper was generated, `protojson` owned serialization, and the int64 came out as a quoted string:

```json
{"reading":{"timestampMs":"1715000000000"}}   // actual
{"reading":{"timestampMs":1715000000000}}     // expected
```

**Root cause.** `collectWrapperMessages` qualified a nested field by membership in a name set built from `collectInt64EncodingContext(file)` — messages declared *in the file being generated*, with a *direct* NUMBER field. Imported types were never in that set (cross-file miss), and wrapper messages were never fed back into it (depth > 1 miss, even within one file).

## The fix

`messageTransitivelyHasInt64Number` walks `field.Message` — which protogen resolves across files — to unbounded depth, with a visited-set cycle guard. This mirrors the existing `messageTransitivelyHasCustomEnum` in `enum_field_encoding.go`, which the `enum_value` generator has used all along and which is why that feature never had this bug.

The predicate being file-independent let the `directMsgNames` parameter be retired rather than patched. `nestsInt64NumberMessage` is widened in lockstep, so the enum conflict check cannot go stale and let a newly-detected wrapper slip through.

## Three further defects surfaced

Each would have left the fix non-functional in practice:

- **`strconv` imported and not used.** A wrapper-only encoding file — exactly what cross-file nesting produces — never uses `strconv`, so the generated code did not compile. Previously unreachable. Golden tests cannot catch this, since they never compile the output; the integration test did.
- **go-client skipped the encoding file for service-less protos.** The message carrying the annotation routinely lives in a pure types file imported by the file defining the RPC — that is exactly the shape in the issue report. `generateInt64EncodingFile` now runs above the services guard, matching go-http and the six sibling generators in the same function.
- **go-client's transitive `UnmarshalJSONSebuf` had no repeated branch**, though its marshal side did, so it tried to decode a JSON array into a single message. Pre-existing; the new fixtures made it reachable.

## Collision handling

Wider detection widens the set of messages that can collide with another `MarshalJSON`-owning annotation (`flatten`, `oneof_config`, `bytes_encoding`, `nullable`, `empty_behavior`, `timestamp_format`, `enum_value`). That combination previously surfaced as a duplicate-method compile error in generated code.

`checkInt64WrapperMarshalJSONConflict` now rejects it at generation time, following the existing `checkEnumMarshalJSONConflict` precedent:

```
message GetSensorReadingResponse: nested int64_encoding=NUMBER requires MarshalJSON but
conflicts with flatten (also requires MarshalJSON) -- only one MarshalJSON-generating
feature is supported per message
```

> [!IMPORTANT]
> This makes generation **newly fail** for those combinations rather than emitting code that does not compile. Loud instead of quiet, but it is a behavior change worth a release note.

## Testing

- Cross-file (`SensorReading` in an imported file, singular + repeated) and depth > 1 (`Outer → Middle → Leaf`, plus a self-referential `Node` exercising the cycle guard) fixtures, wired into **both** Go generators' golden suites. `extraProtoFiles` support added to clientgen's harness.
- Conflict fixture driven through both generators in-process, with a positive control so the rejection cannot over-match.
- **Integration test that compiles the generated code** and asserts the actual JSON bytes for singular and repeated, round-trips both, and pins a control that raw `protojson` still quotes the value — so the numeric output is provably the wrapper's doing.
- `TestGoGeneratorsProduceIdenticalInt64Encoding` now compares **five** goldens instead of one. Comparing only the direct-field golden is precisely what let the two generators drift apart on the wrapper emitters.

**No existing golden changed** — verified against a freshly built binary before regenerating. Full suite passes; `scripts/run_tests.sh` reports 12/12 packages over the coverage gate. Within httpgen/clientgen this introduces no new lint findings and clears one pre-existing `funlen`.

## Out of scope

Adjacent gaps found while investigating, worth separate issues:

- `bytes_encoding`, `timestamp_format`, `nullable`, `empty_behavior` have **no** parent propagation at all — their collectors recurse over nested type *declarations*, not nested field *types*, so a parent silently bypasses the child's marshaler.
- `unwrap`'s `hasEncodingMarshalJSON` remains a depth-0, int64-only check, so the unmarshal side keeps the same depth > 1 blind spot. The marshal side is immune thanks to `emitInlineMarshalChild`; there is no symmetric `emitInlineUnmarshalChild`.


SebastienMelki:
Review findings:

1. P1: Direct NUMBER messages still do not propagate nested NUMBER fields

`internal/httpgen/encoding.go` and `internal/clientgen/encoding.go` skip wrapper collection as soon as a message has any direct `int64_encoding=NUMBER` field. The direct marshaler then only patches `ctx.NumberFields` and never reserializes nested message fields. The new `Node` fixture shows this shape: `Node` has `id` and `child`. `Node{Id:1, Child:&Node{Id:2}}` will emit numeric root `id`, but `child.id` remains quoted because the base `protojson` output is only patched for direct fields. This leaves the “any depth” fix incomplete for recursive or mixed direct+nested messages.

2. P2: Map-value reachability is detected but cannot be emitted

`messageTransitivelyHasInt64Number` counts `map<..., message>` values during the transitive walk, but wrapper field collection/emission only handles singular/repeated non-map fields. For `Parent -> Child -> map<string, Leaf(NUMBER)>`, `Parent` gets a wrapper because `Child` transitively reaches `Leaf`, but `Child` gets no wrapper, so `Parent` falls back to `protojson` for `Child` and the map values still quote the int64. This also risks false conflict errors for messages that are marked as int64 wrappers through a map path the generator cannot actually serialize. Either map values need emitter support or the reachability predicate should reject/ignore that unsupported path consistently.

I did not find other blockers in the PR diff.

Verification: `go test ./internal/httpgen ./internal/clientgen` passes when run with network/local-listener permissions. The sandboxed first run failed on `proxy.golang.org` DNS and `httptest` listener restrictions, not on code failures.

SebastienMelki:
Both findings confirmed and fixed in 8a816df. Thanks — P1 in particular made the "any depth" claim in the PR description false.

## P1 — direct-NUMBER messages now propagate to children

Confirmed exactly as described. `collectWrapperMessages` skipped any message with direct NUMBER fields, and `generateInt64MarshalJSON` only patched `ctx.NumberFields`, so a child that reached an annotated field stayed owned by `protojson` in the base output.

`Int64EncodingContext` now also carries `NestedFields`, and the direct marshal/unmarshal emit the nested re-serialization after patching their own fields. The per-field nested emission is extracted into `emitNestedFieldsMarshal` / `emitNestedFieldsUnmarshal`, shared by the direct and wrapper paths so the two cannot drift apart again.

`Node.MarshalJSONSebuf` now patches `id` **and** recurses into `child`. Only the `Node` golden changed — no other fixture has both direct NUMBER fields and a nested field reaching one.

I verified the new test actually catches this rather than merely passing: reverting just the `emitNestedFieldsMarshal` call in the direct marshaler makes it fail with the precise symptom you described.

```
crossfile_int64_test.go:134: a message with direct NUMBER fields did not propagate to its
nested children: an id is still a quoted string.
--- FAIL: TestMixedDirectAndNestedInt64SerializesAsNumber
```

## P2 — map reachability no longer counted

Also confirmed, including the false-conflict risk. `messageTransitivelyHasInt64Number` no longer descends through map values, so the predicate matches what the emitters can actually traverse. I picked this over adding map emitter support to keep the PR scoped; the limitation is now documented rather than silently half-implemented.

New `int64_map_path_only.proto` pins the behaviour — `Parent -> Child -> map<string, Leaf(NUMBER)>` where `Parent` also carries `flatten`. Generation must succeed and emit no wrapper. Driven through both Go generators, so neither can regress alone.

## Also in this commit

- Integration test now generates the deep-nesting proto into its own package (differing `go_package` values cannot share an output directory) and adds runtime coverage for the `Outer -> Middle -> Leaf` chain plus the mixed `Node` shape, marshal and round-trip. 8 runtime assertions, all executing.
- Extracting the shared emitters dropped all five `//nolint:funlen` directives in these files, including one that predated this branch.

## Verification

Full suite green; `scripts/run_tests.sh` reports 12/12 packages over the coverage gate. All five int64 encoding goldens remain byte-identical between go-http and go-client after normalization. Within `httpgen`/`clientgen` this now sits one finding *below* the pre-PR lint baseline (4 vs 5, all pre-existing `goconst`).

On your sandbox note — the `proxy.golang.org` DNS and `httptest` listener failures were environmental. `TestCrossFileInt64EncodingIntegration` writes a temp Go module and runs `go mod tidy`, so it needs module resolution; `TestMarshalOptionsIntegration` additionally binds a local listener. Both skip cleanly when `protoc` is absent but not when the network is restricted, which is worth knowing for anyone reviewing in a sandbox.

SebastienMelki:
Follow-up review on latest head `8a816df`:

Both prior findings look addressed. Direct `int64_encoding=NUMBER` messages now carry `NestedFields`, and the direct marshal/unmarshal paths call the shared nested emitters. Map-value reachability is no longer counted, the limitation is documented, and `int64_map_path_only.proto` pins the false-conflict case. The generated `Node` golden now reserializes `child`, covering the mixed direct+nested case.

I found no new blockers in the latest update.

Verification run:
- `go test ./internal/httpgen ./internal/clientgen` passes with network/local-listener permissions.
- `go test ./internal/httpgen -run 'TestGoGeneratorsProduceIdenticalInt64Encoding|TestInt64WrapperMarshalJSONConflict|TestCrossFileInt64EncodingIntegration'` passes.\n- `./scripts/run_tests.sh --fast` passes all 12 packages.

PR #226, opened 2026-07-30
fix(openapiv3): map Google well-known wrapper types to scalar OpenAPI schemas

### LINKED ISSUE #225: openapiv3: map Google well-known wrapper types to their scalar OpenAPI schemas
## Problem

When a proto field uses a Google well-known wrapper type (e.g. `google.protobuf.DoubleValue`), the OpenAPI generator treats it as a regular message — emitting a `$ref` to a nested object schema with a `value` property.

```yaml
# Current (incorrect)
myField:
  $ref: '#/components/schemas/DoubleValue'

# where DoubleValue is:
DoubleValue:
  type: object
  properties:
    value:
      type: number
      format: double
```

This is wrong. Per the [protobuf JSON mapping spec](https://protobuf.dev/programming-guides/proto3/#json), wrapper types serialize as their underlying scalar value, not as `{"value": ...}`. `protojson` (used by the Go server at runtime) serializes `DoubleValue{value: 3.14}` as just `3.14`.

The generated OpenAPI spec documents a shape that doesn't match what the server actually returns — breaking clients, codegen tools (openapi-generator), and Swagger UI.

## Expected behavior

Wrapper types should map to their canonical protojson scalar equivalents:

| Wrapper Type | OpenAPI type | format | notes |
|---|---|---|---|
| `google.protobuf.DoubleValue` | `number` | `double` | |
| `google.protobuf.FloatValue` | `number` | `float` | |
| `google.protobuf.Int64Value` | `string` | `int64` | protojson default (string) |
| `google.protobuf.UInt64Value` | `string` | `uint64` | protojson default (string) |
| `google.protobuf.Int32Value` | `integer` | `int32` | |
| `google.protobuf.UInt32Value` | `integer` | `int32` | `minimum: 0` |
| `google.protobuf.BoolValue` | `boolean` | — | |
| `google.protobuf.StringValue` | `string` | — | |
| `google.protobuf.BytesValue` | `string` | `byte` | standard base64 |

Nullability is contextual:
- **Singular wrapper fields** → `type: [<scalar>, "null"]` (the whole point of wrappers is null vs zero-value)
- **Repeated wrapper items** → scalar type only, no null (protojson disallows `null` inside repeated fields)

`int64_encoding` and `bytes_encoding` annotations are **out of scope** — `ValidateBytesEncodingAnnotation()` rejects non-`BytesKind` fields today, and the Go runtime doesn't apply `int64_encoding` to wrapper types. Supporting custom annotations on wrappers is a separate cross-generator alignment issue.

## Context

- The Go server/client generators already handle this correctly via `protojson` at runtime — no changes needed there.
- The Python client generator already handles this correctly with explicit `wellKnownPythonType()` mapping (`pyclientgen/constants.go`, `pyclientgen/types.go`).
- The TypeScript generators also have this problem but are out of scope for this issue.

## Existing pattern

`google.protobuf.Timestamp` is already handled specially in the OpenAPI generator (`types.go:185-187`) via `annotations.IsTimestampField()`. Wrapper types should follow the same pattern — detect by full name, emit the correct scalar schema.

## Implementation notes

This is a two-file change, not just a `types.go` switch addition:

### `internal/openapiv3/types.go`
- Add wrapper type detection at the `MessageKind` case (alongside the existing Timestamp check)
- Emit inline scalar schemas instead of `$ref`s
- Singular wrappers: `type: [<scalar>, "null"]`; repeated wrapper items: scalar only
- `UInt32Value` carries `minimum: 0` (matching existing unsigned scalar conversion)

### `internal/openapiv3/generator.go`
- Skip wrapper types in `collectMessageRecursive()`/`processMessage()` so no wrapper object schemas appear in `components/schemas`
- Handle wrapper types at all RPC-level `$ref` construction sites — request bodies (`line ~902`), responses (`line ~814`), and SSE `x-sse-event-schema` (`line ~932`). Once wrapper components are suppressed, these paths must use the shared resolver or they emit dangling `$ref`s. This includes direct wrapper RPC inputs/outputs (e.g. `rpc GetValue() returns (google.protobuf.DoubleValue)`), which are uncommon but must not produce broken specs.

## Acceptance criteria

- All 9 wrapper types emit inline scalar schemas (not `$ref`s to object schemas)
- Singular wrappers use `type: [<scalar>, "null"]`
- Repeated wrapper items use scalar type only (no null)
- `UInt32Value` retains `minimum: 0` constraint
- No wrapper object schemas appear in `components/schemas` (per-service and bundle mode)
- Direct wrapper RPC inputs, outputs, and SSE event types produce valid schemas with no dangling `$ref`s
- Field comments, examples, and applicable validation constraints are preserved

## Test coverage

Golden test fixtures covering:
- All 9 singular wrapper types
- Repeated wrapper fields
- Map with wrapper values (tested against actual `protojson` behavior, not inheriting singular-field nullability implicitly)
- Direct wrapper RPC input/output
- Component absence verification
- Bundle-mode coverage (component collection and schema naming differ there)

SebastienMelki:
## Summary

- Map all 9 Google well-known wrapper types (`DoubleValue`, `FloatValue`, `Int32Value`, `UInt32Value`, `Int64Value`, `UInt64Value`, `BoolValue`, `StringValue`, `BytesValue`) to their canonical protojson scalar equivalents instead of emitting `$ref` to object schemas with a `value` property
- Singular wrapper fields use `type: [<scalar>, "null"]` (inherent nullability), repeated items and map values use plain scalar (protojson disallows null in both)
- Suppress wrapper object schemas from `components/schemas`, handle direct wrapper RPC inputs/outputs and SSE streaming with inline schemas (no dangling `$ref`s)
- Preserve `field_examples` on wrapper fields

Closes #225

## Test plan

- [x] Golden file tests for `WrapperTypeService` (all 9 singular wrappers, repeated wrappers, map wrappers, direct wrapper RPC input/output, field_examples) in YAML + JSON
- [x] Golden file tests for `WrapperSSEService` (SSE streaming with wrapper output type) in YAML + JSON
- [x] `TestExhaustiveRegression` updated with wrapper_types.proto entries
- [x] All existing OpenAPI golden tests pass (no regressions)
- [x] `golangci-lint` passes with 0 issues
- [x] Component absence verified: no `DoubleValue`, `StringValue`, etc. object schemas in `components/schemas`


internal/openapiv3/types.go
```diff
@@ -332,6 +342,12 @@ func (g *Generator) getMapValueSchema(field *protogen.Field) *base.DynamicValue[
 
 	// Normal scalar or message type
 	valueSchema := g.convertScalarField(valueField)
+
+	// Wrapper type map values are nullable (protojson allows null values in map entries)
```
SebastienMelki: **[P1] Keep map wrapper values non-nullable**

`protojson` does not allow `null` wrapper values inside maps. I verified a nil `map<string, google.protobuf.StringValue>` entry marshals as `""`, while unmarshalling `{"labels":{"nil":null}}` fails with `invalid value for string field value: null`. Adding `"null"` here therefore documents payloads the server rejects. Map wrapper values should use the bare scalar schema, like repeated items; please add a runtime-backed regression test for this case.

internal/openapiv3/types.go
```diff
@@ -186,6 +192,10 @@ func (g *Generator) convertScalarField(field *protogen.Field) *base.SchemaProxy
 		if annotations.IsTimestampField(field) {
 			return g.convertTimestampField(field, schema)
 		}
+		// Handle google.protobuf.* wrapper types as inline scalars
+		if annotations.IsWrapperField(field) {
```
SebastienMelki, line 196: **[P2] Preserve `field_examples` on wrapper fields**

This early return bypasses the common example-decoration block later in `convertScalarField`. I generated a `StringValue` field with two `(sebuf.http.field_examples)` values, and the schema contained only `type: [string, null]`—neither `example` nor `examples` was emitted. That misses issue #225’s explicit requirement to preserve examples. Please route wrappers through the shared decorators (or apply the same example logic in `convertWrapperField`) and cover it in the fixture.

SebastienMelki (commented):
Thanks for covering wrapper mappings across fields, direct RPC schemas, SSE, and component collection. I found two correctness gaps against issue #225’s acceptance criteria: map wrapper values are documented as nullable even though `protojson` rejects null map values, and wrapper fields bypass `field_examples` decoration. Please address the two inline findings before merging.

SebastienMelki (commented):
Re-review complete: both previous findings are addressed. Map wrapper values now use non-null scalar schemas, and wrapper `field_examples` are preserved with golden coverage. `go test ./internal/openapiv3` passes, and all current GitHub checks pass. No remaining code findings.

Non-blocking housekeeping: the PR summary still says “map values use nullable scalar”; that bullet should be updated to match the corrected implementation.

